PR #28, opened 2015-05-15
Explicitly supply the model source

ujamer:
So that `reorderItems` action handler knows which model it is modifying.
Updated code examples in README.

per #22

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -94,7 +94,8 @@ export default Component.extend({
   */
   commit() {
     let items = this.get('sortedItems');
-    let models = items.mapBy('model');
+    let groupModel = this.get('model');
+    let itemModels = items.mapBy('model');
 
     delete this._itemPosition;
 
@@ -111,7 +112,11 @@ export default Component.extend({
         items.invoke('thaw');
       });
     });
-
-    this.sendAction('onChange', models);
+    
+    if  (groupModel === null || groupModel === undefined) {
+      this.sendAction('onChange', itemModels);
+    } else {
+      this.sendAction('onChange', groupModel, itemModels);
+    }
   }
 });
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -2,16 +2,23 @@ import Ember from 'ember';
 import layout from '../templates/components/sortable-group';
 const { $, A, Component, computed, get, set, run } = Ember;
 const a = A;
+const NO_MODEL = {};
 
 export default Component.extend({
   layout: layout,
+
+  /**
+    @property model
+    @type Any
+    @default null
+  */
+  model: NO_MODEL,
 
   /**
     @property items
     @type Ember.NativeArray
   */
   items: computed(() => { return a(); }),
-
   /**
     Vertical position for the first item.
 
@@ -113,10 +120,10 @@ export default Component.extend({
       });
     });
     
-    if  (groupModel === null || groupModel === undefined) {
-      this.sendAction('onChange', itemModels);
-    } else {
+    if (groupModel !== NO_MODEL) {
       this.sendAction('onChange', groupModel, itemModels);
+    } else {
+      this.sendAction('onChange', itemModels);
     }
   }
 });
--- a/tests/dummy/app/routes/application.js
+++ b/tests/dummy/app/routes/application.js
@@ -10,8 +10,8 @@ export default Ember.Route.extend({
   },
 
   actions: {
-    update(newOrder) {
-      this.set('currentModel.items', a(newOrder));
+    update(model, newOrder) {
+      a(model).set('items', a(newOrder));
     }
   }
 });
--- a/tests/dummy/app/routes/application.js
+++ b/tests/dummy/app/routes/application.js
@@ -1,5 +1,5 @@
 import Ember from 'ember';
-const { A } = Ember;
+const { A, set } = Ember;
 const a = A;
 
 export default Ember.Route.extend({
@@ -11,7 +11,7 @@ export default Ember.Route.extend({
 
   actions: {
     update(model, newOrder) {
-      a(model).set('items', a(newOrder));
+      set(model,'items', a(newOrder));
     }
   }
 });
--- a/tests/unit/mixins/sortable-item-test.js
+++ b/tests/unit/mixins/sortable-item-test.js
@@ -103,8 +103,9 @@ test('registers itself with group', function(assert) {
 });
 
 test('deregisters itself when removed', function(assert) {
-  subject.remove();
-
+  run(() => {
+    subject.remove();
+  });
   assert.equal(group.item, undefined,
     'expected to be deregistered with group');
 });
--- a/README.md
+++ b/README.md
@@ -23,8 +23,7 @@ to see if your target browsers are compatible.
 $ ember install ember-sortable
 ```
 
-## Usage
-
+## Basic usage:
 ```js
 // app/routes/my-route.js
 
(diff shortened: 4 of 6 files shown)


ujamer:
Need to work on test cases....

jgwhite:
This is great! Thanks so much :D

One thing we need to consider is that people will have existing action handlers that expect to receive `items` as the first argument. This change would break that expectation.

Here's what I propose:

When `model` is set on `sortable-group`, the `onChange` action is sent with two arguments: `groupModel` and `itemModels`.

Otherwise the behaviour is the same as v1.0.0: `onChange` I sent with only `itemModels`.

How does that sound to you?

ujamer:
sure. Gimme a few minutes to make the change.

tests/dummy/app/routes/application.js
```diff
@@ -10,8 +10,8 @@ export default Ember.Route.extend({
   },
 
   actions: {
-    update(newOrder) {
-      this.set('currentModel.items', a(newOrder));
+    update(model, newOrder) {
+      a(model).set('items', a(newOrder));
```
jgwhite: Why do we wrap the model with `a` here?

tests/unit/mixins/sortable-item-test.js
```diff
@@ -103,8 +103,9 @@ test('registers itself with group', function(assert) {
 });
 
 test('deregisters itself when removed', function(assert) {
-  subject.remove();
-
+  run(() => {
+    subject.remove();
+  });
```
jgwhite: Was this test failing without the runloop? It doesn’t seem like the patch would require this change.

jgwhite:
This is really brilliant work.

Regarding the potential mis-spelling of group’s `model` attribute:

That’s a very good point. We should assume that if the user is using the `model` attr on `sortable-group` then they expect to receive the two argument form of the action.

Perhaps we could give `sortable-group`’s model a sentinel value to detect whether or not it has been set? This would also give us the opportunity to inline-document the property.

```js
const NO_MODEL = {};

export default Component.extend({

  // ...snip...

  /**
    @property model
    @type Any
    @default null
  */
  model: NO_MODEL,

  // ...snip...

  commit() {
    let groupModel = this.get('model');

    // ...snip...

    if (groupModel !== NO_MODEL) {
      this.sendAction('onChange', groupModel, itemModels);
    } else {
      this.sendAction('onChange', itemModels);
    }
  }
});
```

This may be achievable more simply with `undefined`. What do you think?

tests/dummy/app/routes/application.js
ujamer: I'm getting an error on that line when I do not wrap the model in an Ember.Array. It's probably not the best way to fix it, but the error wasn't very descriptive:
```
update@http://localhost:4200/assets/dummy.js:150:9
triggerEvent@http://localhost:4200/assets/vendor.js:35712:13
trigger@http://localhost:4200/assets/vendor.js:59036:9
Router.prototype.trigger@http://localhost:4200/assets/vendor.js:57760:9
EmberRouter<.send@http://localhost:4200/assets/vendor.js:35210:8
ActionHandler<.send@http://localhost:4200/assets/vendor.js:38901:9
TargetActionSupport<.triggerAction@http://localhost:4200/assets/vendor.js:42089:17
Component<.sendAction@http://localhost:4200/assets/vendor.js:50729:1
commit@http://localhost:4200/assets/vendor.js:75866:9
_tellGroup@http://localhost:4200/assets/vendor.js:76136:9
_complete@http://localhost:4200/assets/vendor.js:76221:7
Backburner.prototype.join@http://localhost:4200/assets/vendor.js:10799:18
run.join@http://localhost:4200/assets/vendor.js:27952:12
run.bind/<@http://localhost:4200/assets/vendor.js:28014:14
tryCatch@http://localhost:4200/assets/vendor.js:59478:16
invokeCallback@http://localhost:4200/assets/vendor.js:59490:17
publish@http://localhost:4200/assets/vendor.js:59461:11
@http://localhost:4200/assets/vendor.js:38429:7
Queue.prototype.invoke@http://localhost:4200/assets/vendor.js:11432:11
Queue.prototype.flush@http://localhost:4200/assets/vendor.js:11497:13
DeferredActionQueues.prototype.flush@http://localhost:4200/assets/vendor.js:11302:13
Backburner.prototype.end@http://localhost:4200/assets/vendor.js:10727:11
Backburner.prototype.run@http://localhost:4200/assets/vendor.js:10782:15
executeTimers@http://localhost:4200/assets/vendor.js:11164:7
updateLaterTimer/backburner._laterTimer<@http://localhost:4200/assets/vendor.js:11153:11
```

tests/unit/mixins/sortable-item-test.js
ujamer: Yeah, I was running into that failing test on my end. Not in the scope of this patch though. Let me revert that. Sorry!

ujamer:
Thanks! It's been a good learning experience for me too. :)

Let me give that a spin and see.

tests/unit/mixins/sortable-item-test.js
```diff
@@ -104,7 +104,6 @@ test('registers itself with group', function(assert) {
 
 test('deregisters itself when removed', function(assert) {
   subject.remove();
```
jgwhite: You know, looking at this now, I think you were completely right to wrap this in a `run`. I’m surprised this test was ever working! Could you reinstate the `run`? Sorry!

tests/dummy/app/routes/application.js
jgwhite: Ah… this is because `model` is a POJO rather than an `Ember.Object`. I guess the simplest thing is to use `Ember.set`:

```js
const { A, set } = Ember;

update(model, newOrder) {
  set(model, 'items', a(newOrder));
}
```

addon/components/sortable-group.js
```diff
@@ -111,7 +118,11 @@ export default Component.extend({
         items.invoke('thaw');
       });
     });
-
-    this.sendAction('onChange', models);
+    
+    if (groupModel === null) {
```
jgwhite: I’m not sure we can compare with `null` because it’s possible the object passed as `model` could be `null` too.

Let’s say the rule is:

If you write `{{#sortable-group model=anything ...}}` in your template then you get the two argument action, no matter what `anything` evaluates to.

tests/unit/mixins/sortable-item-test.js
ujamer: No worries, added it back.

addon/components/sortable-group.js
ujamer: Well, I'm assuming that if the object passed as `model` is null, there shouldn't be any items in the object for it to invoke `commit()`.  

Otherwise, we'd need to pre-define `model` as something other than `null`, because that's what it's using to check if user has passed in something `undefined`.

jgwhite: My fear is that a `null` model may arrive due to an unrelated error but it’ll be unclear what’s gone wrong when it gets to the action handler because rather than a `null` argument in the first position you’ll get a list of items.

The [suggestion above](https://github.com/jgwhite/ember-sortable/pull/28#issuecomment-102829337) uses a unique sentinal value:

```js
const NO_MODEL = {};
```

This is guaranteed not to collide with any other value `model` might have.

jgwhite:
Going to call it a night but will pick up the thread in the moring. This is super close to being ready to merge :)

tests/dummy/app/routes/application.js
ujamer: fixed.

addon/components/sortable-group.js
ujamer: Ah, missed the const definition when reading over it, so I was scratching my head on the definition of `NO_MODEL`, and assumed that it's a null equivalent. That should do the trick. Let me update it.

ujamer:
Should be good. Have a good night!

jgwhite:
Perfect. LGTM

ujamer:
Awesome!

PR #31, opened 2015-06-03
New Computed Syntax

jmurphyau:
This is to implement the new computed property syntax to removes deprecation warning from Ember 1.13.

If you are Ok with only supporting Ember v1.12.0-beta.1+ then the use of a plugin can be removed - let me know if you would prefer to go that way and I'll update the PR.

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import layout from '../templates/components/sortable-group';
-const { $, A, Component, computed, get, set, run } = Ember;
+import computed from 'ember-new-computed';
+const { $, A, Component, get, set, run } = Ember;
 const a = A;
 const NO_MODEL = {};
 
@@ -18,29 +19,35 @@ export default Component.extend({
     @property items
     @type Ember.NativeArray
   */
-  items: computed(() => { return a(); }),
+  items: computed({
+    get() { return a(); }
+  }),
   /**
     Vertical position for the first item.
 
     @property itemPosition
     @type Number
   */
-  itemPosition: computed(function() {
-    let element = this.element;
-    let stooge = $('<span style="position: absolute" />');
-    let result = stooge.prependTo(element).position().top;
+  itemPosition: computed({
+    get() {
+      let element = this.element;
+      let stooge = $('<span style="position: absolute" />');
+      let result = stooge.prependTo(element).position().top;
 
-    stooge.remove();
+      stooge.remove();
 
-    return result;
+      return result;
+    }
   }).volatile(),
 
   /**
     @property sortedItems
     @type Array
   */
-  sortedItems: computed('items.@each.y', function() {
-    return a(this.get('items')).sortBy('y');
+  sortedItems: computed('items.@each.y', {
+    get() {
+      return a(this.get('items')).sortBy('y');
+    }
   }),
 
   /**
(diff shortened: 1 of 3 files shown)


addon/components/sortable-group.js
```diff
@@ -18,29 +19,35 @@ export default Component.extend({
     @property items
     @type Ember.NativeArray
   */
-  items: computed(() => { return a(); }),
+  items: computed({
+    get() { return a(); }
```
jgwhite: I was under the impression that `computed(function() {})` and `computed({ get() {} })` were equivalent with the new form.

/cc @cibernox

package.json
```diff
@@ -30,6 +30,7 @@
     "ember-data": "1.0.0-beta.16.1",
     "ember-export-application-global": "^1.0.2",
     "ember-disable-prototype-extensions": "^1.0.0",
+    "ember-new-computed": "1.0.0",
```
jgwhite: Nice. I wasn’t aware this existed.

jgwhite:
Thanks for these contributions and sorry I’ve been so slow to respond! :grimacing:

jgwhite:
That one query aside, this looks good to merge. Once it’s in, I’ll upgrade ember-cli and cut a new release.

addon/components/sortable-group.js
jmurphyau: Nah.. Computed properties that are passed a function (which means one function for get and set) throw deprecation warnings in newer versions of Embet

jgwhite: Okay, cool. In that case, let’s get this merged :D

jmurphyau: Yeah that was the whole trigger to me creating PRs - I had deprecation warnings in my app about computed properties I wanted to get rid of.. 

Started with this one then thought 'why not add demoURL too' - since u already had the site as a demo.. :)

cibernox: This change should not be necessary.

`computed(function() {})` is just syntax sugar to `computed({ get() {} })`, and shouldn't throw any deprecation warning.

jmurphyau: I'll go back to my app and confirm where the warning came from - but if definitely was related to computed properties that are provided a function rather than an object. 

I believe it was in the `y` computed property..

cibernox: **Maybe** the deprecation warning had something to do with calling `set` on computed property with no explicit set defined.

jmurphyau: https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/computed.js#L121

jgwhite: That’s only if the `arity > 1` though, right? In this case it’s 0.

cibernox: Exactly. What it's deprecated are CPs like this:

```
computed(function(key, settedValue) {  
  if (arguments.length > 1) {
    // setter logic
  } else {
    // getter logic
  }
})
```

jmurphyau: Ahh ok. I didn't know that

The deprecation was caused by this line:

`y: computed(function(_, value) {`

And since I was updating that one I figured may as wel update all of them.. Wasn't aware it was the argument length that caused the deprecation..

Good to know.. That will change the way I do some of my computed properties in the future.

jgwhite: Definitely good knowledge. I suggest we stick with the explicit `get` form in this PR anyways.

cibernox: Indeed, there is no harm in using the explicit `get` apart from a few extra keystrokes.

PR #35, opened 2015-07-06
Use `closest` vs `is` to support complex handles (e.g. SVG)

rlivsey:
Ran into this because we're using SVGs for our icons & it didn't recognise that the handle was being dragged. This uses `closest` instead just to check that we're dragging something which is either the handle, or inside the handle.

Couldn't see any existing tests for handles, but added an inner span to the dummy app as this triggers the problem.


jgwhite:
Good spot! Could you rebase?

rlivsey:
Rebased

jgwhite:
Thanks!

Issue #50, opened 2015-08-04
onChange fires onClick


mehulkar:
it looks like when i click an element, it also fires the onChange action. i'm guessing this also reinserts the element? i'm yielding a component templet inside the sortable items and they lose their state onChange. So they're also losing their state

jgwhite:
This can be counteracted today by declaring a `handle`.

I’d be open to a PR that suppress `onChange` unless the order actually changes. It’ll add complexity but sounds valuable.

mehulkar:
I have a `handle` defined, just FYI. 

I'll think about how to do implement, but honestly, more likely I'll end up changing my design to _not_ have an "expand" button in each element that needs to be maintained. Super spike project and this library got me 80% of the way, so I'm satisfied for now. (Feel free to close this, if you don't think the additional complexity is a huge win)

jgwhite:
@mehulkar I’ll keep this open. I think it’s something we probably *should* account for.

opsb:
@jgwhite Vice versa when you drag an item the click event fires... there doesn't appear to be an easy way to tease the two apart. jquery-ui-sortable takes the approach of only initiating a drag if the mouse is moved after the mousedown. It prevents the click after a successful drag by adding a listener to catch the next click.

jgwhite:
Closed by #51

PR #51, opened 2015-08-09
Distinguish between clicking and dragging an item

opsb:
PR to solve https://github.com/jgwhite/ember-sortable/issues/50. 

This delays the start of a drag until the mouse is moved. If a drag is completed it also prevents a click from being fired on the item that's just been dragged.

I'm afraid there's no tests included, not sure how to test drag n drop. I can circle back but I need to take a look at denzo's fork https://github.com/denzo/ember-sortable and get dragging between lists working within my app as a first priority (currently swapping out an existing component that used jquery-ui-sortable). 

For manual testing I've introduced an {{item-presenter item=item}} component, which reports 'Item clicked' to the console when it's clicked, and added it to the horizontal demo. When an item is dragged you'll see 'Updated'. You should see that they're now mutually exclusive.

I've tested in IE(9), chrome(latest), safari(latest) and firefox(latest).


opsb:
Possible library for testing drag events https://github.com/bitovi/syn

jgwhite:
This is perfect, thanks so much. I like the approach you took with verifying behaviour. Now that component integration tests have arrived we'll be using them in the suite.

jgwhite:
Going to test this out when I'm back in front of a machine but the code looks spot on. Could you squash to one commit?

opsb:
@jgwhite Squashed. There's a little simplification that makes use of bind, I'm guessing this is reasonable as IE9+ supports bind and IE8 doesn't support 2D transforms. However I can back this out if you can imagine a scenario where someone might use a polyfill to add the 2D transforms.

jgwhite:
Given that Ember 2.0 will no longer support IE8, I’m happy to proceed with `bind`.

jgwhite:
Released as v1.5.1

mehulkar:
beautiful people. thanks @opsb!

PR #58, opened 2015-09-01
Support scrolling while dragging

eugeniodepalo:
This PR adds support for scrolling the list container (or any other parent) while the mouse is dragging an item. It does so by remembering what was the offset of the element relative to the document when the drag starts, and by calculating the new offset on every mouse move.

## Before

p

## After

c


tests/acceptance/smoke-test.js
```diff
@@ -67,6 +70,36 @@ test('reordering with mouse events', function(assert) {
...
+  };
+
+  drag(
+    'mouse',
+    '.scrollable-demo .handle[data-item=Uno]',
+    () => {
+      return { y: itemHeight() + 1 };
+    },
+    {
+      dragend: function() {
```
eugeniodepalo: At first I tried adding these sort of callbacks to the `reorder` helper, but that didn't really work because this test is has to rely on the offset to make sense, and `reorder` works by targeting the elements directly.

jgwhite:
Awesome. Will review tonight.

addon/mixins/sortable-item.js
```diff
@@ -308,14 +308,18 @@ export default Mixin.create({
...
 
     if (groupDirection === 'x') {
       dragOrigin = getX(startEvent);
       elementOrigin = this.get('x');
+      scrollOrigin = $(this.element).offset().left;
 
       return event => {
         let dx = getX(event) - dragOrigin;
-        let x = elementOrigin + dx;
+        let elementX = this.get('x') - this.element.offsetLeft;
```
jgwhite: Is `this.element.offsetLeft` different to `$(this.element).offset().left`?

addon/mixins/sortable-item.js
```diff
@@ -308,14 +308,18 @@ export default Mixin.create({
...
+        let scrollX = $(this.element).offset().left - elementX;
+        let x = elementOrigin + dx + (scrollOrigin - scrollX);
```
jgwhite: Could you walk me through what’s happening on lines 320–322? I can’t quite picture it :bowtie:

addon/mixins/sortable-item.js
```diff
@@ -431,17 +438,16 @@ export default Mixin.create({
...
+      return new Promise(resolve => {
+        run.next(() => {
+          let duration = this.get('transitionDuration');
+          run.later(this, resolve, duration);
+        });
       });
-    });
+    } else {
+      return Promise.resolve();
+    }
```
jgwhite: Remind me why we need this conditional… was it to make integration testing easier?

tests/helpers/drag.js
```diff
@@ -0,0 +1,82 @@
...
+  let start, move, end;
+
+  if (mode === 'mouse') {
+    start = 'mousedown';
+    move = 'mousemove';
+    end = 'mouseup';
+  } else if (mode === 'touch') {
+    start = 'mousedown';
+    move = 'mousemove';
+    end = 'mouseup';
```
jgwhite: I just realised that I messed up my reorder helper and never used the touchevent names!

jgwhite: I have fixed this on master in reorder. These should be `touchstart`, `touchmove`, `touchend`.

tests/helpers/drag.js
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+  Drags elements by an offset specified in pixels.
+
+  Examples
+
+      drag(
+        'mouse',
+        '.some-list li[data-item=uno]',
+        { y: 50 }
```
jgwhite: The third argument should be a function, right?

tests/helpers/drag.js
```diff
@@ -0,0 +1,82 @@
...
+  andThen(() => {
+    let item = findWithAssert(itemSelector);
+    let itemOffset = item.offset();
+    let offset = offsetFn();
+
+    triggerEvent(item, start, {
+      pageX: itemOffset.left,
+      pageY: itemOffset.top
+    }).then(() => {
+      if (callbacks.dragstart != null) {
```
jgwhite: No need to `!= null`.

tests/helpers/drag.js
```diff
@@ -0,0 +1,82 @@
...
+        callbacks.dragstart();
+      }
+    });
+
+    triggerEvent(item, move, {
+      pageX: itemOffset.left,
+      pageY: itemOffset.top
+    }).then(() => {
+      if (callbacks.dragmove != null) {
```
jgwhite: No need to `!= null`.

tests/helpers/drag.js
```diff
@@ -0,0 +1,82 @@
...
+    triggerEvent(item, move, {
+      pageX: itemOffset.left + offset.x,
+      pageY: itemOffset.top + offset.y
+    });
+
+    triggerEvent(item, end, {
+      pageX: itemOffset.left + offset.x,
+      pageY: itemOffset.top + offset.y
+    }).then(() => {
+      if (callbacks.dragend != null) {
```
jgwhite: No need to `!= null`.

jgwhite:
@eugeniodepalo the `drag` test helper is a really good idea. I wonder if we can implement `reorder` in terms of `drag`?

addon/mixins/sortable-item.js
jgwhite: The `x` computed property uses scroll offset (line 120). Can the usage be unified? Should it be applied to the `y` computed property also?

addon/mixins/sortable-item.js
eugeniodepalo: Yes, `offsetLeft` returns the distance relative to the offset parent, and will also exclude transforms, while `$(element).offset().left` returns the distance relative to the document.
So by subtracting `this.element.offsetLeft` from `this.get('x')` you basically get the transform offset. I agree that this is kind of a hack and hard to follow - I think I should instead store the original position somewhere every time we set `x`/`y`.

I'm doing this because in order to find the difference in scroll position, the reference point must be the same. If I just do `scrollOrigin - $(this.element).offset().left`, this is not going to yield the scroll offset, because the `left` property will also include the transform previously set to the element.

I'm really not sure if there are better ways to do this - would gladly appreciate any input you have :)

addon/mixins/sortable-item.js
eugeniodepalo: Here we basically account for the scroll to determine the real `dx`, instead of just relying on the distance from the mouse starting point (which will not change if you scroll an element that is parent of the list). So the actual `dx` is the mouse offset + the difference in scroll from when you started dragging.

Once again, not entirely sure this is the best approach!

addon/mixins/sortable-item.js
eugeniodepalo: Yes, without this the integration "unit test" wasn't working, because that kind of test doesn't actually wait for promises to solve. I think I'm free to remove this now that we switched to a proper acceptance test.

tests/helpers/drag.js
eugeniodepalo: Woops - good catch :)

tests/helpers/drag.js
eugeniodepalo: Ahah I didn't review it while copying it - thanks will rebase again!

eugeniodepalo:
> I wonder if we can implement reorder in terms of drag?

Totally agree - I'll try to refactor `reorder` as a special case of drag.

addon/mixins/sortable-item.js
jgwhite: Gotcha. I don’t know if there’s a better approach either. Maybe let’s just add some comments to illuminate?

jgwhite:
I’m happy we’re taking time over this PR. Lots of solid improvements are coming from it.

eugeniodepalo:
@jgwhite I refactored the reorder helper to use drag, how does it look?

I've also simplified the logic that calculates the scroll offset by just using the parent element so that the element transform doesn't mess with our numbers. Do you feel it's easier to understand now?

jgwhite:
Looking very cool indeed. I’m going to play around with this a little on my machine to make sure I understand all the moving parts but I think we’re pretty much ready.

mehulkar:
Is this good to merge?

jgwhite:
@mehulkar not yet. I’m going to do a thorough review at tonight’s Ember London project night. Want to be sure I really understand the code before merging.

jgwhite:
Reviewed and ready.

mehulkar:
do i have to do anything special to make use of this? I was using `v1.5.3` and just updated to `1.6.2`, but the page doesn't scroll when I try to drag a list item further below the page. Not sure what to expect.

jgwhite:
@mehulkar this patch does not enable automatic scrolling, but rather makes ember-sortable aware of scroll offsets. An important piece of prerequisite work to get smart scrolling implemented.

Issue #54, opened 2015-08-25
emit mouseDown event


mehulkar:
In a long-ish list, when re-ordering, I want to zoom out the UI (apply a CSS transform to shrink all the elements).
This will allow the user to easily drag and drop elements a large number of elements away.
I think I could do this if we had a mouseDown event and i could do custom stuff from it. 

What do you think? I think it would mean exposing another handle like `onChange`.

jgwhite:
Seems reasonable to me. The UI you describe sounds cool. I wonder if the hook you’re after is something like `prepare` or `before-sort` or `will-sort` — something that can return a promise so can defer sorting until animation is complete.

mehulkar:
hmm. maybe I'm just unfamiliar with how the code is written, but I would have assumed that the "sorting" part is triggered when the element is dropped, no? In my head, I'm seeing it like this:

1. mouseDown
1. custom hook
1. dragEvent
1. dropEvent
1. onChange
1. custom hook

Soemthing like `will-sort` seems like it would fit in after `dropEvent` or after `onChange`. 
I'm looking for something after `mouseDown` and also after `dropEvent` to reverse whatever I did on mouseDown. 

Does that make sense?

jgwhite:
I think you’re right that `will-sort` implies completing the sort.

Could you sketch out how you’d use this if it were implemented as `mouseDown`?

jgwhite:
Mainly I’m curious as to what component will receive the action in this particular example.

mehulkar:
I just looked through the code a little bit and it looks like the `onChange` event is fired in sortable-group in the `commit` function. I think `mouseDown` could be fired in the `prepare` function? A quick spike validated that I was able to send the action pretty easily!

jgwhite:
@mehulkar I would still like to explore to possibility of giving this action a more specific name. Three reasons:

1. It could be a `touchstart` that triggers it
2. We’re cancelling the real `mousedown` with `preventDefault`
3. This may open the door to exposing many other DOM events as actions

It might not be a problem to overload `mousedown` with our synthetic version but I’d like to explore other possibilities.

This hook definitely has value but let’s make sure we nail it.

mehulkar:
One thing that's tripping me up looking at the code is that I don't
understand what the sortable groups are for. Could you talk about a use
case for that?

On Monday, August 31, 2015, Jamie White <notifications@github.com> wrote:

> @mehulkar <https://github.com/mehulkar> I would still like to explore to
[...]

PR #60, opened 2015-09-03
Add actions for `onDragStart` and `onDragStop`

drapergeek:
I had the need to change a property in my parent template when something was dragging vs when it wasn't. Following the "Data Down, Actions Up", this seemed like the cleanest implementation.

I don't care for testing private methods but I didn't think it warranted a full on event and such for just testing these actions.

I also removed a `console.log` that didn't seem to add any benefits. If there was a reason for that, just let me know.

Happy to take feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,37 @@ slightly different colour:
 }
 ```
 
+# Drag Actions
+
+The `onDragStart` and `onDragStop` actions are available for the
+`sortable-item`s. You can provide an action name to listen to these actions to
+be notified when an item is being dragged or not.
+
+When the action is called, the item's model will be provided as the only
+argument.
+
+```js
+// app/routes/my-route.js
+
+export default Ember.Route.extend({
+  actions: {
+    dragStarted(item) {
+      console.log(`Item started dragging: ${item.get('name')`);
+    },
+    dragStopped(item) {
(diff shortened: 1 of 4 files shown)


jgwhite:
Very nice. I believe this pretty much answers @mehulkar’s issue #54 too. You’re right about the `console.log`, not sure what that was for.

jgwhite:
This looks good to merge. Could you add documentation of these actions to the README? Ultimately I’d like to show off this feature in the demo but that can be a separate PR.

drapergeek:
I just pushed up the changes to add documentation to the README. Hope that does the job.

I also started adding the demo but can't find a good way to demo it without adding a controller or a component because you can't access the properties on the route in the template. I'm guessing we'd prefer to wait for routable components for that instead or would you prefer a controller? I'm happy to write  it, won't take much time at all.

jgwhite:
Released in 1.6.0 :sunglasses:

drapergeek:
Thanks!

mehulkar:
Nice work! This isn't the `mouseDown` event in #54, but I think it works for the use case I was talking about. @jgwhite feel free to close #54 if you don't think another mouseDown is required.

jgwhite:
@mehulkar awesome, thanks!

PR #61, opened 2015-09-18
v1.6.1

jgwhite:


ammorrise:
@jgwhite Question for you about this change. It allows you to scroll with the mouse while you drag, but is there an option to have it scroll automatically when you hit the bottom of the screen?

jgwhite:
@ammorrise no, that feature has not landed yet. I don’t belive anyone is actively working on it so it’s certainly up for grabs.

PR #83, opened 2016-02-04
Fix test imports usage in README

seanpdoyle:
As a consumer of the `test-support` helpers, importing in
`tests/test-helper.js` causes issues.

To resolve these issues, recommend that consumers import the helpers as
part of `tests/helpers/start-app.js` instead.


seanpdoyle:
This is what we recommend with [ralphs-little-helpers](https://github.com/thoughtbot/ralphs-little-helpers#usage).

jgwhite:
Perfect, thanks :)

PR #87, opened 2016-02-23
A new sortable-item property: spacing

dianafa:
This PR introduces a new property to ember-sortable: the spacing which appears after dragging an element. When users starts to drag element, other elements jump back, which is a nice visual effect. It works both for the x and y option;
In `y` case elements above current one jump up, and elements below current one - jump down.
In `x` case elements before current one jump to the left, and elements after current one - jump to the right.
The `spacing` is defined for each item which uses the `sortable-item` mixin (and can be different for each of the elements).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@
 /libpeerconnection.log
 npm-debug.log
 testem.log
+.idea
--- a/README.md
+++ b/README.md
@@ -94,6 +94,19 @@ To change sort direction, define `direction` on `sortable-group` (default is `y`
 {{#sortable-group direction="x" onChange="reorderItems" as |group|}}
 ```
 
+### Changing spacing between currently dragged element and the rest of the group
+
+When user starts to drag element, other elements jump back. Works both for the `x` and `y` direction option.
+
+In `y` case: elements above current one jump up, and elements below current one - jump down.
+In `x` case: elements before current one jump to the left, and elements after current one - jump to the right.
+
+To change this property, define `spacing` on `sortable-item` (default is `0`):
+
+```hbs
+{{#sortable-item tagName="li" group=group spacing=15}}
+```
+
 ### CSS, Animation
 
 Sortable items can be in one of three states: default, dragging, dropping.
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -30,12 +30,14 @@ export default Component.extend({
 
   /**
     Position for the first item.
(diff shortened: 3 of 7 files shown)


jgwhite:
Thanks for this PR and thanks for using the addon :D

Reviewing quickly now.

dianafa:
Cool! Thanks!

jgwhite:
I 100% agree that this is a good feature to add. However, I wonder if the spacing could be declared entirely with CSS?

So, rather than declaring

```hbs
{{sortable-group spacing=50}}
```

We’d say:

```css
.sortable-item.is-dragging {
  margin-top: 50px;
  margin-bottom: 50px;
}
```

After some experimentation it appears that ember-sortable doesn’t handle the appearance of this extra margin completely gracefully. If you agree that this might be a good way to go then I believe the logic in this patch can be put to work implementing this CSS solution.

What do you think?

dianafa:
Yes, it doesn't :( I've tried to add this margins to `is-busy` class (after adding `classNameBindings: ['isBusy'],` to an item) but still it's not what would satisfy me (and other potential users of this addon), as in case of more complex sortable elements, adding additional margins is not enough and actually breaks styling.

I find the solution with `spacing` property more clean and straightforward. It doesn't require additional work from the user side and I think it's nothing bad that despite in some cases it can be obtained via CSS, we expose this property as a customizable option. How does that sound? :)

jgwhite:
> I find the solution with spacing property more clean and straightforward. It doesn't require additional work from the user side and I think it's nothing bad that despite in some cases it can be obtained via CSS, we expose this property as a customizable option. How does that sound? :)

Sounds good to me :)

Could you add documentation of this option to the README and, if you have time, apply it to one of the examples on the demo page?

dianafa:
Hi, README updated, as well as demo page code. Btw, why is this in the `tests/dummy/` directory? :)

jgwhite:
Awesome :D

> why is this in the tests/dummy/ directory?

It’s a convenient convention with Ember addons. That page is also the target for acceptance tests.

This all looks great. Many, many thanks. Would you mind squashing into one commit? Then I think we’re good to merge.

dianafa:
thanks!

done :)

jgwhite:
Let’s get this merged! :grin:

jgwhite:
Released in 1.8.0

dianafa:
Yay! Great, thank you for cooperation and so fast action :)

PR #95, opened 2016-05-11
Fixes Internet Explorer and Edge transform issue - Closes #94

### LINKED ISSUE #94: In Internet Explorer and Edge, elements with display type of "table-row" do not have the "ghosting" effect during drag.
In Internet Explorer and Edge, elements with display type of "table-row" do not have the "ghosting" effect during drag (which makes it hard for a user to see where they are dragging/dropping elements to/from). This problem can be seen in the table example on the ember-sortable demo page. I have looked in to the issue and have found [this relevant stackoverlfow question](http://stackoverflow.com/questions/34718443/ie-edge-not-applying-transform-translate-to-table-row/34730094#34730094). It refers to the fact that in IE and Edge elements with the display type "table-row" do not support the transform css property, even though the specification implies that both browsers do. However, as the SO question mentions, the elements with the display type "table-cell" work just fine. 

I have implemented a solution to adapt to this IE and Edge bug in my own project by extending ember-sortable. It needs to be refined for general use, but if my solution is of interest, I would love to adapt it and submit a PR. 

Here is what I am using now, and what I think can be adapted into a solution for ember-sortable:

``` javascript
import Ember from 'ember';
import SortableItemMixin from 'ember-sortable/mixins/sortable-item';

export default Ember.Mixin.create(SortableItemMixin, {
    transformableElement: null,

    didInsertElement() {
        this.set('transformableElement', this._checkIfTransformableElement());
        return this._super(...arguments);
    },

    freeze() {
        let transformableElement = this.get('transformableElement');
        if (!transformableElement) {
            let cells = this.$('td');
            if (!cells) { return; }

            cells.css({ transition: 'none' });
            cells.height(); // Force-apply styles
        }

        return this._super(...arguments);
    },

    reset() {
        let transformableElement = this.get('transformableElement');
        if (!transformableElement) {
            let cells = this.$('td');
            if (!cells) { return; }

            cells.css({ transform: '' });
            cells.height(); // Force-apply styles
        }

        return this._super(...arguments);
    },

    thaw() {
        let transformableElement = this.get('transformableElement');
        if (!transformableElement) {
            let cells = this.$('td');
            if (!cells) { return; }

            cells.css({ transition: '' });
            cells.height(); // Force-apply styles
        }

        return this._super(...arguments);
    },

    _applyPosition() {
        let transformableElement =  this.get('transformableElement');
        if (!transformableElement) {
            let el = this.$();
            if (!el) { return; }

            let cells = this.$('td');
            if (!cells) { return; }

            const groupDirection = this.get('group.direction');

            if (groupDirection === 'x') {
                let x = this.get('x');
                let dx = x - el.prop('offsetLeft') + parseFloat(el.css('margin-left'));
                let dChildX = dx;

                cells.css({
                    transform: `translateX(${dChildX}px)`
                });
            }

            if (groupDirection === 'y') {
                let y = this.get('y');
                let dy = y - el.prop('offsetTop');
                let dChildY = dy;

                cells.css({
                    transform: `translateY(${dChildY}px)`
                });
            }
        }

        return this._super(...arguments);
    },

    _checkIfTransformableElement(displayType) {
        displayType = displayType || this.$().css('display');

        const scenarios = [
            { display: "table-row", selector: "tr" },
            { display: "table-cell", selector: "td" },
            { display: "table-row-group", selector: "tbody" },
            { display: "table-caption", selector: "caption" },
            { display: "table-header-group", selector: "thead" },
            { display: "table-footer-group", selector: "tfoot" }
        ];

        let testTable = Ember.$(
            `<table id="test-drag-support">
                <caption></caption>
                <thead>
                    <tr><th></th></tr>
                </thead>
                <tbody>
                    <tr><td></td></tr>
                </tbody>
                <tfoot>
                    <tr><td></td></tr>
                </tfoot>
            </table>')`
        );

        // insert the test table so we can perform calculations
        testTable.css({
            display: 'block',
            position: 'absolute',
            top: '-1000px'
        });
        testTable.appendTo('body');

        let pass = true;

        for (let i = 0; i < scenarios.length; ++i) {
            let scenario = scenarios[i];

            if (displayType.toLowerCase() === scenario.display) {
                /* Test Transformability */
                let element = testTable.find(scenario.selector).first(),
                    initial = element[0].getBoundingClientRect().top, 
                    transformed;

                element.css({ transform: 'translateY(10px)'});
                transformed = element[0].getBoundingClientRect().top;

                pass = transformed == initial + 10;
                break;
            }
        }

        // remove the test table
        testTable.remove();

        return pass;
    }
});
```

The idea for the code which checks the transformability is straight from [this JSFiddle](https://jsfiddle.net/jonathansampson/9tog66gn/) which was linked to in the SO post I mentioned.

cojomojo:
This fixes #94 with a mixin that can be used as a user of the addon sees fit. By using a separate mixin, users that do not need the patch do not have to include it. The fix involves inserting a temporary table off screen and running tests live after the `didInsertElement` event. This is the only way, as far as I can tell, to verify if an element supports the transform type. I included the same unit tests which are used for sortable-item, and then an additional test to check the core element testing logic. Checkout [my instance of the demo](http://cojomojo.github.com/ember-sortable/demo) (the table example) based off this fix to see what it really provides.


cojomojo:
Not sure why the CI failed. There seems to be something phantomjs 1.9 doesn't like, because all tests pass with 2.1 and when tests are ran in chrome with `ember test -s`.

jgwhite:
Hi, sorry for the extremely slow response to this! I got behind on notifications and am only just catching up.

Thoughts:

1. Amazing work!
2. `_checkIfTransformableElement` seems quite expensive, given it’s called for every item…
   1. Can we make it lazier?
   2. Is there any way to do this work just once and share the results between all items?
3. It’s a pain that the test is failing in Phantom 1.9 but…
   1. Does that mean there’s a potential edge case lurking out there?
   2. Given this is an IE-specific fix, should this test even run in Phantom?
   3. Should I set up AppVeyor? (yes)

cojomojo:
No problem on the response speed. 

I agree, `_checkIfTransformableElement` is expensive. I'm not sure what you have in mind to make it 'lazier', but we could most definitely make it run less often by storing the results from a check, and doing a lookup to that value before running the function. For example, if the element is tr, and we already have a stored result from  `_checkIfTransformableElement` for a tr then we don't call `_checkIfTransformableElement` again. This would cut the cost quite a bit. I'll go ahead and do this for sure. If we cut the expense of the fix down, perhaps it would be better to integrate the fix into the sortable-item mix in instead of having an additional fix mixin. Thoughts?

As for the test failing in phantom 1.9... I will look into it a bit more to make sure I am not missing something. AppVeyor sounds like a good idea to me too.

jgwhite:
The strategy you describe sounds ideal and general purpose enough that it could indeed go directly into sortable-item

jgwhite:
@cojomojo are you still interested in pursuing this patch? If not, do you mind if I close this PR?

cojomojo:
@jgwhite We can go ahead and close this, I don't think it will be a problem for 99% use case, and it is a bit of a heavy change for the 1%. Also, I no longer need the patch, so all is well on my end.

jgwhite:
Many thanks @cojomojo

PR #100, opened 2016-07-18
Fix for dragging outside of ember-modal-dialog boundaries

davelowensohn:
A simple fix to prevent errors when sortables are dragged outside the boundaries of a containing `{{ember-modal-dialog}}` component.


jgwhite:
Hi @davelowensohn. Massive massive apologies this has gone so long without review or reply. I have been asleep at the wheel a little bit with this project 😞 

This PR looks good to go. Will merge and cut a release now.

Issue #105, opened 2016-10-21
The reorder helper no longer works in ember-cli-qunit 3.0


omghax:
It seems as though the changes made to `ember-cli-qunit` in ember-cli/ember-cli-qunit#48 have broken the `reorder` test helper's calculations. The dragged items only move half the distance they're supposed to, and so the list never actually changes.

jgwhite:
Good spot, and awesome fix.

PR #106, opened 2016-10-22
Fix test helpers in ember-cli-qunit 3.0

omghax:
This is a fix for #105. I've updated the `drag` helper to apply a scaling factor to `dx` and `dy`, which fixes the failing tests when running under `ember-cli-qunit` 3.0, and should remain compatible with 2.x as well.

### DIFF
--- a/addon/helpers/drag.js
+++ b/addon/helpers/drag.js
@@ -52,8 +52,9 @@ export function drag(app, mode, itemSelector, offsetFn, callbacks = {}) {
     let item = findWithAssert(itemSelector);
     let itemOffset = item.offset();
     let offset = offsetFn();
-    let rect = item.get(0).getBoundingClientRect();
-    let scale = item.outerHeight() / (rect.bottom - rect.top);
+    let itemElement = item.get(0);
+    let rect = itemElement.getBoundingClientRect();
+    let scale = itemElement.clientHeight / (rect.bottom - rect.top);
     let targetX = itemOffset.left + offset.dx * scale;
     let targetY = itemOffset.top + offset.dy * scale;
 
--- a/addon/helpers/drag.js
+++ b/addon/helpers/drag.js
@@ -52,8 +52,10 @@ export function drag(app, mode, itemSelector, offsetFn, callbacks = {}) {
     let item = findWithAssert(itemSelector);
     let itemOffset = item.offset();
     let offset = offsetFn();
-    let targetX = itemOffset.left + offset.dx;
-    let targetY = itemOffset.top + offset.dy;
+    let rect = item.get(0).getBoundingClientRect();
+    let scale = item.outerHeight() / (rect.bottom - rect.top);
+    let targetX = itemOffset.left + offset.dx * scale;
+    let targetY = itemOffset.top + offset.dy * scale;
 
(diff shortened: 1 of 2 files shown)


jgwhite:
Many thanks for the PR. I’ve just been looking at the same set of problems but your fix is way better than mine (`#ember-testing { transform: none !important }` 😝).

Unfortunately, there is a gnarlier issue — the mixin unit tests are fundamentally incompatible with the next release of Ember. We could delete them, or port them to a component integration test.

I’m going to think through the options and either merge this PR as it stands or cherry pick the change into another PR that fixes the fundamental issue.

If you’re feeling heroic and can see a clear way to resolve the mixin test problems then please be my guest 😉.

acburdine:
@omghax would you be willing to rebase this branch against the latest master? I fixed the test suite so that should make the tests pass.

I also merged @jgwhite's solution to this problem (didn't realize that was included with the update pr 😬) so would you be willing to remove that change with this PR? (relevant lines: https://github.com/jgwhite/ember-sortable/blob/master/tests/dummy/app/styles/app.css#L158-L161)

omghax:
@acburdine Thanks for fixing the tests :) I just rebased it and added a commit to remove the `#ember-testing` patch.

acburdine:
@omghax awesome, thanks :)

omghax:
@acburdine which tests are failing for you? I tried it locally with Chrome 54 and everything passed :\

acburdine:
@omghax half of the acceptance tests failed - as well as the mixin unit tests

omghax:
@acburdine I've noticed mousing over the browser while the tests are running will cause Ember to throw runloop errors. Could it be that? Otherwise, I'm not sure why they'd pass for me but not for you.

acburdine:
@omghax Ah, that would make sense then about the acceptance tests - however I'm still getting errors with the mixin tests getting incorrect height & width.

acburdine:
@omghax for better reference, here's the actual errors I'm getting:

```sh
not ok 88 Firefox 49.0 - component:sortable-item-mixin: height
    ---
        actual: >
            35
        expected: >
            60
        stack: >
            @http://localhost:7357/assets/tests.js:783:5
            runTest@http://localhost:7357/assets/test-support.js:2998:14
            Test.prototype.run@http://localhost:7357/assets/test-support.js:2983:4
            run/<@http://localhost:7357/assets/test-support.js:3146:6
            process@http://localhost:7357/assets/test-support.js:2822:4
            begin@http://localhost:7357/assets/test-support.js:2804:2
            internalStart/config.timeout<@http://localhost:7357/assets/test-support.js:3574:4

        message: >
            expected height to be height + margin-bottom
        Log: |
    ...
not ok 89 Firefox 49.0 - component:sortable-item-mixin: width
    ---
        actual: >
            45
        expected: >
            70
```

Same issue in Chrome and Firefox - not getting the issue in phantomjs though

acburdine:
@omghax Sorry for all the spam - I did figure out the issue

The mixin unit tests are failing because the css fix is still needed, as the component unit tests don't use the drag helper. So in the end, we'll keep both changes, esp. as the drag helper can be used in composing app test suites 😄

PR #112, opened 2016-10-26
Remove event listeners in willDestroyElement

### LINKED ISSUE #110: Memory leak
[sortable-mixin](https://github.com/jgwhite/ember-sortable/blob/master/addon/mixins/sortable-item.js#L299) doesn't clean up its `primeDrag` method every time, resulting in ember-sortable retaining memory.

The leak happens specifically when a drag action is started, but never completed.

We're seeing ~9Mb of memory being retained in our test suite by this method not being cleaned up completely.

tim-evans:
`cancelStartDragListener` is retaining components in tests due to the event listener not always being torn down. When I was tracking down memory leaks in our application, we found that this method was one of the points in our test suite that would retain a bunch of memory (~9Mb or so) through our test run of 1,000 tests.

This fixes #110


acburdine:
Looks good, thanks!

PR #114, opened 2016-10-28
Replace sendAction with ember-invoke-action

### LINKED ISSUE #85: Ensure hooks work with closure actions
Not sure if they do or not, but would be good to check.

acburdine:
closes #85
- add ember-invoke-action dependency
- replace sendAction usage

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -1,6 +1,8 @@
 import Ember from 'ember';
 import layout from '../templates/components/sortable-group';
 import computed from 'ember-new-computed';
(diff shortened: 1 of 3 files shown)


mcfiredrill:
Just curious why was this chosen over normal closure actions?

acburdine:
@mcfiredrill I don’t believe closure actions were supported in all supported versions of ember at the time this was made. Could have been wrong at the time, although regardless of the reason I think it would be fine standardizing on closures now

mcfiredrill:
@acburdine Thanks for your reply!

>  I think it would be fine standardizing on closures now

Would you be interested in a PR that does this?

Issue #115, opened 2016-11-03
Performance


estshy:
This component has low performance (especially on mobile) because the touchdown and mousedown handlers are neither debounced nor throttled. In addition events are actually attached to each element of the list instead of being delegated to sortable group. 

To make it possible to attach mousedown and allow people to drag items, dragging should be default prevented only after moving it by let's say 10px otherwise custom events handlers inside sortable-item are broken on mobile (touchstart has higher precedence than click and touchdown).

scottmessinger:
We'd love to see a PR for this! If you need any help, let us know!

jgwhite:
Agreed. All the proposed improvements sound great and I would love to see them come to fruition. Perhaps we could do smart stuff with the capture phase too?

estshy:
@scottmessinger  @jgwhite I've improved the performance of dragging (https://github.com/jgwhite/ember-sortable/pull/118). I will try to focus on the remaining issues when I will find more time.

jgwhite:
@estshy awesome, many thanks. Someone will review your PR soon.

scottmessinger:
Since #118 has been merged, I'm going to close this for now. Thanks @estshy for the PR!!!

PR #118, opened 2016-11-11
Dragging performance improvment

estshy:
Improving performance of dragging by throttling drag event using ember run loop. It should be a nice entry for the list of performance issues (https://github.com/jgwhite/ember-sortable/issues/115).

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -3,6 +3,7 @@ import computed from 'ember-new-computed';
 import scrollParent from '../system/scroll-parent';
 import ScrollContainer from '../system/scroll-container';
 import {invokeAction} from 'ember-invoke-action';
+import { throttle } from 'ember-runloop';
 
 const { Mixin, $, run } = Ember;
 const { Promise } = Ember.RSVP;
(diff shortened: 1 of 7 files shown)


jgwhite:
Thanks for the PR.

This seems to introduce changes that aren’t related to event throttling. For example: changing formatting and removing the auto-scrolling logic.

Could you reduce this to just the necessary code changes?

estshy:
Sure, there you have.

acburdine:
@estshy thanks for the PR! I'll get it looked at a bit later today :)

addon/mixins/sortable-item.js
```diff
@@ -320,17 +321,18 @@ export default Mixin.create({
     if (this.get('isBusy')) { return; }
 
     let drag = this._makeDragHandler(event);
+    let dragThrottled = ev => throttle(this, drag, ev, 50, false);
```
scottmessinger: If we want 60fps, we want this to be `1000/60 = 16`. `33` if we want 30fps. Having twiddled with settings like this before, I'm not sure what the best number is. For really complex interactions where each frame takes a long time to draw (50+ms), I've found lowering the frame rate keeps animations smoother. So, ideally, I'd love if we could find an easy way to customize this. If not, I vote for `16` instead of `50` so we're targeting 60fps

estshy: @scottmessinger I think we can also refresh the coordinates (x and y) every 50-100 milliseconds and make smooth animation using CSS transitions. In that case the delay would be visible only if the user would have change direction of dragging.

acburdine:
Going to go ahead and release 1.9.1 with the autoscrolling bugfix - will look at getting this into a 1.10 release unless @scottmessinger thinks it's good to go?

estshy:
@acburdine @scottmessinger What's the status?

scottmessinger:
LGTM! @acburdine I'm good to merge if you are!

PR #120, opened 2016-11-14
adds support for 'data-test-selector' attribute

cspanring:
Adds (optional) attribute bindings for `data-test-selector` to the `sortable-group` and `sortable-item` component. 

Not a critical change, rather an improvement for test ergonomics in consuming applications, by following a widely adopted convention around testing with page objects.

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -10,6 +10,8 @@ const NO_MODEL = {};
 export default Component.extend({
   layout: layout,
 
+  attributeBindings: ['data-test-selector'],
+
   /**
     @property direction
     @type string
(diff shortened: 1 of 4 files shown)


jgwhite:
Seems like a good idea to me. What do you reckon @acburdine?

cspanring:
Thanks. To be clear: feel free to close this PR if you don't think it's within the scope of the sortable components. One can add the attribute bindings with a `reopen` in the consuming application. It's just not as user friendly as if the component would support that convention out of the box.

acburdine:
@cspanring SGTM 👍 

Had to look up what the test selector was all about (hadn't used them before), but now it makes sense 😄 

Thanks for the PR!

PR #126, opened 2017-01-05
Dragging threshold

estshy:
Right now dragging starts on mousedown and touchstart which means it basically starts on click which is wrong. I am constantly moving elements on mobile when I am scrolling and toggling things. There should be dragging threshold (minimal number of pixels which element needs to be moved in order to be considered dragged).

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -2,6 +2,8 @@ import Ember from 'ember';
 import computed from 'ember-new-computed';
 import scrollParent from '../system/scroll-parent';
 import ScrollContainer from '../system/scroll-container';
+import {invokeAction} from 'ember-invoke-action';
+
 const { Mixin, $, run } = Ember;
 const { Promise } = Ember.RSVP;
 
@@ -333,7 +335,7 @@ export default Mixin.create({
 
     this._tellGroup('prepare');
     this.set('isDragging', true);
-    this.sendAction('onDragStart', this.get('model'));
+    invokeAction(this, 'onDragStart', this.get('model'));
 
     this._scrollOnEdges(drag);
   },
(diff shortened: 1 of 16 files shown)


addon/mixins/sortable-item.js
```diff
@@ -119,6 +119,16 @@ export default Mixin.create({
    * @public
    */
   maxScrollSpeed: 20,
+  
+  /**
+   * How many pixes the element needs to be moved to be considered dragged
```
lukemelia: `s/pixes/pixels/`

addon/mixins/sortable-item.js
```diff
@@ -119,6 +119,16 @@ export default Mixin.create({
...
+   *
+   * @property draggingThreshold
+   * @type Number
+   * @default 15
+   * @public
+   */
+  draggingThreshold: 15,
```
lukemelia: I think the default should probably be `1`

lukemelia:
We've run into similar needs and were hoping something like this was available in ember-sortable.

addon/mixins/sortable-item.js
estshy: @lukemelia I guess it's hard to make a click without moving the mouse even by 1 pixels. I played with different values and that one came out as a good compromise.

In my option the value should block accidentally dragging by clicking for most of the situations  by default.

acburdine:
@estshy This LGTM, and I will try to get this tested either today or tomorrow. Fixing @lukemelia's catch with the typo's probably not a bad idea though 😉

addon/mixins/sortable-item.js
estshy: @lukemelia Changed. Thanks.

acburdine:
@estshy any chance of getting a rebase and some changes re: the above?

mupkoo:
Some other libraries have `delay` option. Maybe this is something that can be added instead of the `draggingThreshold`?

acburdine:
Forgot about this - closed by #137 as it fixes the issue present in this PR.

PR #127, opened 2017-01-13
publicize supported ember versions

kellyselden:
Going by your ember-try testing on Travis, this is your minimum supported ember version. Is it correct, and do you want a badge for it?


acburdine:
Thanks @kellyselden!

PR #128, opened 2017-01-14
remove unused ember-new-computed polyfill

kellyselden:
As a result of #127, it appears you do not need this polyfill anymore, as it targets ember versions prior to 1.12.0-beta.1.


jgwhite:
LGTM. What do you think @acburdine ?

acburdine:
I think it's good - although this technically constitutes a breaking change - so will release this with ember-sortable 2.0.

jgwhite:
@acburdine very true. Good call.

PR #135, opened 2017-03-22
Dramatically increase performance by reducing relayout

scottmessinger:
The combination of calling `css` and then `height` as in `node.$().css(...).height()` was causing a recalculate styles -> relayout for **each** slot. In my app, this increased performance of the `clear` function by approximately 1000x (1300ms -> 1.3ms). Granted, my app has a bit over 200 draggable elements, so the difference is more pronounced. In more normal apps, the change should still be noticeable.

### DIFF
--- a/addon/utils/slot.js
+++ b/addon/utils/slot.js
@@ -153,7 +153,7 @@ export default class Slot {
 
     if (node.isDestroyed || node.isDestroying) { return; }
 
-    node.$().css({ width: '', height: '', transform: '' }).height();
+    node.$().css({ width: '', height: '', transform: '' });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


jgwhite:
I’m struggling to remember why that was added in the first place. I assume there must have been some need to forcefully flush the change to DOM before the next step. Still, if it still works then great!

scottmessinger:
@jgwhite I wondered that, too. My other hypothesis was height was originally being returned from the function.

acburdine:
I will get this looked at/tested as soon as possible 👍

jgwhite:
@acburdine note this is a PR against the experimental ja-abrams branch. We should discuss whether/how to resurrect efforts there. 

Are you gents at emberconf?

acburdine:
@jgwhite oops, completely missed that this was on the jj-abrams branch 😛, may take me a little bit longer to get this looked at then.

I will be at emberconf though 👍

jgwhite:
Excellent. We must make time to hang out!

PR #136, opened 2017-03-30
Fix when surrounded by ember link-to

maartenparmentier:
Somehow we needed this when our item was an Ember #link-to.
Without this we couldn't get the sorting to work.

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -673,7 +673,11 @@ export default Mixin.create({
    * @private
    */
   _preventClick(element) {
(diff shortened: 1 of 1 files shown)


mupkoo:
I have this in an app as well, to prevent the event from propagating to the link

```javascript
// app/components/sortable-item.js
import SortableItem from 'ember-sortable/components/sortable-item';
import $ from 'jquery';

export default SortableItem.extend({
  _preventClick(element) {
    $(element).one('click', function (e) {
      e.stopPropagation();
      e.preventDefault();
      e.stopImmediatePropagation();
    });
  }
});
```

jgwhite:
Thank you so much for the patch! We’ll release it later today.

PR #140, opened 2017-06-01
make `_scrollOnEdges()` get `itemContainer` height dinamically

marlosin:
This "fix" is kind of related with the autoscroll feature that was added on `1.8.3`. If the dragged element height is higher than its container, the autoscroll starts moving automatically. This change does not fix that, but allows that overriding element's height (by a css class for example) gets correctly calculated. 

Please let me know if I was clear enough or if you guys need some demo.


jgwhite:
Thank you so much for this patch! We’ll release it later today.

PR #150, opened 2017-08-16
Disable isDragging when element is destroyed

nbibler:
While debugging an issue in an application, we came across a case where the act of dragging an element caused re-renders and actively-dragging-element destruction.

In this scenario, ember-sortable would endlessly scroll the page to the top of the screen because the `isDragging` flag remained true (once dragging began) even after the element was destroyed. This occurred because the `_scrollOnEdges` method uses `requestAnimationFrame` to recursively call itself while dragging. If the element is destroyed and isDragging is forever true, these recursive calls will never end.


jgwhite:
Thank you so much for this submission. I’m really sorry it hasn’t received any attention. I hope to devote the time to reviewing it soon.

PR #151, opened 2017-08-16
Alias sortable direction to group

rmachielse:
We are using this addon in an application where `group` is already a thing in many places.
Therefore we would like to use a different name for that variable.
With this change we only have to override `_direction` and `_tellGroup` instead of having to copy complete methods.

Let me know if this is desired or if you know a better way!

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -234,6 +234,13 @@ export default Mixin.create({
   }).volatile(),
 
   /**
+    @private
+    Allows host instance to use the `group` property for something else with
+    minimal overriding.
+  */
+  _direction: computed.readOnly('group.direction'),
+
+  /**
     @method didInsertElement
   */
   didInsertElement() {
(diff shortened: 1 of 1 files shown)


rmachielse:
@jgwhite @acburdine any thoughts about this?

acburdine:
So I'm fine with this change for the 1.x branch I think - my idea with a 2.0 release (once I get around to cleaning some stuff up 😬) is to refactor how properties are passed between the two components.

rmachielse:
Ok, great! What way are you thinking of? Should I change the pr to a different branch?

rmachielse:
@acburdine can this be merged soon?

rmachielse:
@acburdine is there anything that is keeping this from being merged? Please let me know if you need anything from me

jgwhite:
Sorry we haven’t moved forward on this. I hope to get up to speed on it soon and get back to you with the next step.

jgwhite:
Thanks so much for the patch! I’ll try to release it later today!

rmachielse:
@jgwhite awesome, thanks a lot!

PR #156, opened 2017-10-04
fix for issue with transforms being left after drop

### LINKED ISSUE #132: Position of the items are scattered after the sorting
Some css transform is applied while sorting and it is not removed. I guess it  causes the misbehavior

shidel-dev:
fixes #132. 

`_drag` was being invoked after `reset` was called on the item, and after a drop.

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -580,6 +580,9 @@ export default Mixin.create({
     @private
   */
   _drag(dimension) {
+    if(!this.get("isDragging")) {
+      return;
+    }
     let updateInterval = this.get('updateInterval');
     const groupDirection = this.get('group.direction');
 


shidel-dev:
Not sure if there is a way to have the events stop firing after a drop occurs (it seems to only happen in some cases). I just did some debugging to get my information (placed logs in _drag, and reset).

shidel-dev:
It was also unclear if I should make my changes on the 1.x branch or master. I tried out the master branch in my application, and it seems to have issues with performance, and also it seems as if my "handle" on the element is coming loose. I did not dive into this. perhaps there is an api change I am not aware of.

shidel-dev:
I do have a version of my changes that makes these changes on the mater branch as well.

mupkoo:
I have the same problem and this patch seems to solve it.

jgwhite:
Thanks so much for this patch and sorry it hasn’t received any attention. I can’t make any exact promises about when it’ll be merged but don’t let us off the hook, we’ll get there!

jgwhite:
Thanks so much for the patch! I’ll try to release it later today.

PR #154, opened 2017-09-15
Up/down buttons

H1D:
Adding actions yielded from `{{sortable-item}}` to move up/down using buttons.
sortable buttons

Closes #153.

 - [x] tests
 - [x] implementation
 - [x] demo
 - [x] docs

---

PR #157, opened 2017-10-18
Add tabindex attribute binding

hjdivad:
This enables simple keyboard-based accessible sorting via  a `keyUp` handler.

For example:

```hbs
  {{#each myItems as |item idx| }}
    {{#sortable-item tabindex=0 keyUp=(action 'keyUp' idx) model=item }}
      {{item.name}}
    {{/sortable-item}}
  {{/each}}
```

```js
actions: {
  keyUp(index, evt) {
    switch(evt.key) {
      case "ArrowDown":
        // move item at `index` back one
        break;
      case "ArrowUp":
        // move item at `index` forward one
        break;
      default:
        return;
    }
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,19 @@ No data is mutated by `sortable-group` or `sortable-item`. In the spirit of “d
 
 Each item takes a `model` property. This should be fairly self-explanatory but it’s important to note that it doesn’t do anything with this object besides keeping a reference for later use in `onChange`.
 
+### Accessibility
+
+`sortable-item`s can receive a `tabindex` which allows them to be focused.  Use this to enable keyboard sorting for accessibility.
+
+```hbs
+{{#each myItems as |item idx| }}
+  {{#sortable-item tabindex=0 keyUp=(action 'keyUp' idx) tagName="li" model=item group=group handle=".handle"}}
+    {{item.name}}
+    <span class="handle">&varr;</span>
+  {{/sortable-item}}
+{{/each}}
+```
+
 ## Testing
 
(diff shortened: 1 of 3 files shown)


hjdivad:
This is complementary to #154 which provides a more "batteries included" button-based approach to accessibility.

There's also a [twiddle](https://ember-twiddle.com/11656f42f6d25de11432ee5d9d67db83?openFiles=templates.application.hbs%2Ctemplates.components.x-component.hbs) highlighting the basic idea.

jgwhite:
This looks great to me. The only tiny tweak I’d make is to remove the `onDragStart="dragStarted" onDragStop="dragStopped"` from the example snippet in the readme. Just so readers focuses on the pertinent attribute.

Thanks!

jgwhite:
@hjdivad I took the liberty of rebasing this against 1.x and making the small README change I mentioned. I’ll try to cut a release later today.

hjdivad:
@jgwhite ah i was about to follow up on your excellent suggestion re the README change but I see you've already done it. ^_^

alanwguo:
@jgwhite The example that @hjdivad posted that includes the implementation of keyUp is great. However, the official documentation does not include the keyUp implementation. I was unsure of how tabindex alone allows us to implement sortable-item in an accessible way until I stumbled upon this PR.

I can send a PR to update the documentation to include the example implementation of keyUp in the documentation or perhaps a blurb describing how keyUp should be implemented to support reordering in an accessible way. Do you think that's a good idea?

jgwhite:
@alanwguo that sounds fantastic. Thanks!

PR #158, opened 2017-10-19
ember-cli-update to 2.16.2

jgwhite:
Result of running [`ember-cli-update`](https://github.com/kellyselden/ember-cli-update) on the project.


package.json
```diff
@@ -2,50 +2,49 @@
...
-  "author": "",
-  "license": "MIT",
   "devDependencies": {
     "broccoli-asset-rev": "^2.4.5",
-    "ember-ajax": "^2.4.1",
-    "ember-cli": "2.9.1",
-    "ember-cli-app-version": "^2.0.0",
-    "ember-cli-dependency-checker": "^1.3.0",
-    "ember-cli-htmlbars-inline-precompile": "^0.3.3",
+    "ember-ajax": "^3.0.0",
```
Dhaulagiri: i suspect this is not actually used in this addon

package.json
```diff
@@ -2,50 +2,49 @@
...
     "start": "ember server",
     "test": "ember try:each"
   },
-  "repository": "https://github.com/jgwhite/ember-sortable",
-  "engines": {
-    "node": ">= 0.12.0"
+  "dependencies": {
+    "ember-cli-babel": "^6.6.0",
+    "ember-cli-htmlbars": "^1.0.10",
+    "ember-invoke-action": "^1.4.0",
```
Dhaulagiri: could we bump this to 1.5.0 to get some recent updates including babel 6?

package.json
```diff
@@ -2,50 +2,49 @@
...
+    "ember-new-computed": "^1.0.2"
```
Dhaulagiri: I think this is only for Ember 1.11 and earlier, so I wonder if this could be dropped at this point?

rmachielse:
@jgwhite @acburdine any idea when we can expect this?

Dhaulagiri:
@jgwhite how would you feel about merging this as is and I can make a separate PR to update to 2.18 and address my comments from this one?

package.json
Dhaulagiri: Ignore me here. Consumers can do this already via a yarn upgrade.

jgwhite:
@Dhaulagiri sounds perfect, merging 👌

PR #159, opened 2017-10-31
Update ember-cli and dependencies.

btecu:
Use new ember modules.
Drop bower.


jgwhite:
Thanks so much for the PR! 😀

I think this might double up some of the changes in #158 

Would you mind checking and maybe we can combine our changes?

btecu:
Not sure how I missed that one. The only difference I see is that yours also added the `yarn.lock`. 👍

Issue #143, opened 2017-07-04
Add a configurable OVERSHOOT property for test helper (reorder)


BnitoBzh:
I made different tests with the reorder test helper.
In some case, the reordering works correctly, but in some other cases it doesn't works ..
I tried to update the "OVERSHOOT" value, and where initialy the test failed, with a value to 20 it works.

I think it's a good idea to allow to override the value directly in the reorder function by passing a new attribute.

Or, in order to prevent breaking change, by change the dy and dx calculation like this :

const OVERSHOOT = 2;
let dx = targetElement.offset().left - OVERSHOOT - element.offset().left;
let dy = targetElement.offset().top - OVERSHOOT - element.offset().top;

TO

const OVERSHOOT = 1.1;
let dx = targetElement.offset().left - OVERSHOOT * element.offset().left;
let dy = targetElement.offset().top - OVERSHOOT * element.offset().top;

This change allow a best rounding errors management.

Should i make a pull request ?

BnitoBzh:
const OVERSHOOT = 1.05; is better ...

pboutin:
I had some inconsistencies in my tests as well. Changing the `OVERSHOOT` fixed it for me as well.

I suggest having `reorder(app, itemSelector, resultSelectors, options)` as the new contract (having the overshoot and the mode into an option hash with defaults.

I can open a PR if you want.

jgwhite:
@pboutin a PR would be awesome. Thank you!

jgwhite:
This project is now being maintained by the team at @heroku.

To get the project healthy and maintainable again, we're doing a bit of spring cleaning of older issues and pull requests. This issue hasn't been touched in a while so we're closing it.

Please feel free to re-open if you'd like to resume the conversation.

PR #160, opened 2017-11-15
Fix the reorder test helper

pboutin:
There was a problem with the overshoot when calculating
drag distances. Most of the time, the dragged element
was not dragged enough to cause a proper reordering.

A nice upgrade to this would be to change the contract
and move the `overshoot` and the `mode` into an option
hash.

fix issue #143

### DIFF
--- a/addon/helpers/reorder.js
+++ b/addon/helpers/reorder.js
@@ -5,7 +5,7 @@ import Ember from 'ember';
   To avoid rounding errors, we must therefore double
   the overshoot.
 */
-const OVERSHOOT = 2;
+const OVERSHOOT = 1.05;
 
 /**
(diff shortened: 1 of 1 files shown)


pboutin:
fix issue #143

jgwhite:
@pboutin ah yes — the reason for an overshoot of exactly 2px in the existing tests is because the app is scaled down to 0.5x. A 1px overshoot would tend to get rounded away.

jgwhite:

To get the project healthy and maintainable again, we’re doing a bit of spring cleaning of older issues and pull requests. This pull request hasn’t been touched in a while so we’re closing it.

Please feel free to re-open if you’d like to resume work on it.

PR #162, opened 2017-12-07
Remove jquery event listener from component element

jackson-dean:
I am working removing memory leaks across test runs in a very large ember app. I noticed that jquery seems to maintain a reference to the element here because of the event listener. I thought this was a bit strange as I expected the listener would be eligible for garbage collection once the element is removed from the dom. However, I believe the issue is that jquery maintains a cache of elements passed to its constructor and does not know to update its internal cache if the element is removed using native apis or some other non-jquery technique. The result is when when the test run is complete, there is an application container instance that cannot be garbage collected. This should not affect any logic or functionality. just notifies jquery that the listener should be removed when the component is destroyed.

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -611,7 +611,7 @@ export default Mixin.create({
   _drop() {
     if (!this.element || !this.$()) { return; }
 
-    this._preventClick(this.element);
+    this._preventClick();
 
     this.set('isDragging', false);
     this.set('isDropping', true);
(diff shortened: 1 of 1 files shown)


nlfurniss:
Hey @jgwhite, any concerns with this PR?

nlfurniss:
@jgwhite would it be possible to get a review for this? Would love to fix this memory leak

jgwhite:
Apologies @nlfurniss! I’ll get this reviewed, merged, and released today.

jgwhite:
Release as [v1.11.1](https://github.com/jgwhite/ember-sortable/releases/tag/v1.11.1)

Issue #166, opened 2018-01-16
Consider removing ember-new-computed


Dhaulagiri:
This is only needed (i think) for ember 1.11 and earlier which is _pretty_ old at this point

PR #175, opened 2018-02-23
remove ember-new-computed

Dhaulagiri:
This was a polyfill for ember 1.11 apps which seems ok to remove at this point since I can't think of many reasons someone would be stuck on 1.11 at this point (and the testing matrix for this addon only tests 2.x anyhow)

Closes https://github.com/jgwhite/ember-sortable/issues/166


jgwhite:
Released in [1.11.2](https://github.com/jgwhite/ember-sortable/releases/tag/v1.11.2)

PR #181, opened 2018-05-09
Add modern test helpers

alexlafroscia:
This adds an `async/await`-friendly version of the `drag` and `reorder` test helpers, that compose better with the "modern" test helpers provided by `@ember/test-helpers`.

I'm not sure if the `offsetLeft` and `offsetTop` properties will work in all browsers, but it does work in Chrome. I opted for this approach so that this doesn't depend on `jQuery` in any way. If you'd rather I put that back to the `jQuery`-full approach, I'm happy to do that.

Note: Because they're provided in `addon-test-support` they would be imported like

```javascript
import drag from 'ember-sortable/test-support/drag';
import reorder from 'ember-sortable/test-support/reorder';
```

### DIFF
--- a/addon-test-support/reorder.js
+++ b/addon-test-support/reorder.js
@@ -1,0 +1,21 @@
+import { findAll } from '@ember/test-helpers';
+import drag from './drag';
+
+export default async function reorder(mode, itemSelector, ...resultSelectors) {
+  const items = findAll(itemSelector);
+
+  await resultSelectors.reduce(async (acc, selector, targetIndex) => {
+    await acc;
+
+    const element = items.find(item => item.matches(selector));
+    const targetElement = items[targetIndex];
+    const dx = targetElement.offsetLeft - element.offsetLeft;
+    const dy = targetElement.offsetTop - element.offsetTop;
+
+    if (dx === 0 && dy === 0) {
+      return Promise.resolve();
+    }
+
+    return drag(mode, element, () => ({ dx, dy }));
+  }, Promise.resolve());
+}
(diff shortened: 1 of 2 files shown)


addon-test-support/reorder.js
```diff
@@ -0,0 +1,21 @@
+import { findAll } from '@ember/test-helpers';
+import drag from './drag';
+
+export default async function reorder(mode, itemSelector, ...resultSelectors) {
+  const items = findAll(itemSelector);
```
pgengler: I think this needs to be inside the `reduce`; otherwise it doesn't reflect the changed order after each iteration. (At least, that's something I had to change to make it work in one of my apps.)

mupkoo:
Hello @alexlafroscia 

The native `offsetTop` and `offsetLeft` depend on the `offsetParent` of the node, so a given node can have an `offsetTop` that equals 20, but when you add all the parent offsets, it can easily grow into thousands.

One solutions is to loop over the `offsetParent` or use `getBoundingClientRect`

alecha:
Is there a plan to ship this?

jgwhite:
Apologies — I hadn’t been tracking this effectively. Giving the build a kick.

jgwhite:

To get the project healthy and maintainable again, we're doing a bit of spring cleaning of older issues and pull requests. The work in this PR is super valuable. Once the project is healthy again we’ll look to apply your hard work, even if it’s not specifically this PR.

Please feel free to re-open if you'd like to resume the conversation.

PR #182, opened 2018-07-17
Updated README to include js example for accessibility.

alanwguo:


jgwhite:
Thanks for this. There seems to be a problem that’s preventing build running. I’ll look into it.

PR #185, opened 2018-08-03
docs: use action helper in more examples

knownasilya:


jgwhite:
Thanks for this. Re-running the build

jgwhite:
Hmm, seems to be a problem that’s preventing build running. I’ll look into it.

knownasilya:
👍

PR #188, opened 2018-09-26
Don't break if A extensions are not enabled

btecu:
Addon breaks in Ember 3.4 if Array extensions are disabled


jgwhite:
Thanks for the patch! I need to fix CI but wants it’s back this looks good to merge.

btecu:
This is really a quick patch. Optimally, the addon shouldn't use or rely on those array prototype extensions.
`sortBy('field')` is just `sort(x => x.field)` in native javascript.

PR #212, opened 2018-11-27
Replace jQuery with vanilla JS

nlfurniss:
With Ember moving towards a jQuery-less world, it would be awesome to have this addon continue to work without $.

- Tests pass
- all the example apps function as expected

Note: I changed the MouseEvent property that's being read (previously `pageX/Y` to `clientX/Y`) because ember-test-helpers/ember-native-dom-helpers don't allow that property to be set when creating events in tests. However, there is no change in functionality.


jgwhite:
Thanks so much for the PR, it looks really solid. A member of the team will review it in detail as soon as possible.

fivetanley:
@nlfurniss thanks for the PR! it looks great, my initial question is can we use `@ember/test-helpers` as well? It looks like they support jquery-less mode also.

nlfurniss:
> @nlfurniss thanks for the PR! it looks great, my initial question is can we use `@ember/test-helpers` as well? It looks like they support jquery-less mode also.

I could, but that requires updating the test syntax to the new QUnit syntax, adding babel async/await support, etc and would make this PR very large and difficult to review

jgwhite:
Let’s think about this from the point of view of consuming apps that use the `drag` test helper.

Questions on my mind:

1. Do the helpers from `@ember/test-helpers` not work at all in old-school tests?
2. Are we OK introducing `ember-native-dom-helpers` as a transitive dependency for consuming apps?

Given that the extra dependency will only affect the dev/test build for consuming apps, I think this is a reasonable tradeoff. Nonetheless we should consider its impact and be sure it’s not possible to use `@ember/test-helpers` for the same purpose.

nlfurniss:
Good news - I was very wrong about `test-helpers! I had tried to use them over native-dom-helpers in another project and ran into an issue that must have been specific to that project. Everything works perfectly using test-helpers here.

Please note, I set Babel to not transpile `async/await` since it's only being used in tests running in headless Chrome

nlfurniss:
Spoke too soon 😞

Even though the tests pass locally in headless chrome and in real chrome using `test-helpers`, for some reason in CI the element's `getBoundingClientRect` values seem to be [undefined](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/fire-event.ts#L103) which is causing tests to fail.

jgwhite:
@nlfurniss great detective work!

The thing that jumps out at me from the screenshot is:

image

Where’s the `dx` gone? 😅

This should come from the calling `reorder` helper:

```js
let dx = targetElement.offset().left - OVERSHOOT - element.offset().left;
let dy = targetElement.offset().top - OVERSHOOT - element.offset().top;

drag(mode, element, () => { return { dx: dx, dy: dy }; });
```

Any idea what happened to `dx`?

jgwhite:
@nlfurniss that looks pretty good to me.

nlfurniss:
@jgwhite I added the missing property, as well as upgraded from `ember-cli-qunit` to `ember-qunit` because cli-qunit is deprecated and ember-qunit helped me debug this issue locally

jgwhite:
Thanks for your rigour and fantastic work @nlfurniss!

jgwhite:
@nlfurniss your fine work has been released as part of [1.12.2](https://www.npmjs.com/package/ember-sortable/v/1.12.2). Thanks again 🎩

addon/mixins/sortable-item.js
```diff
@@ -292,49 +297,46 @@ export default Mixin.create({
     @method freeze
   */
   freeze() {
-    let el = this.$();
+    let el = this.element;
     if (!el) { return; }
 
-    el.css({ transition: 'none' });
-    el.height(); // Force-apply styles
+    el.style.transition = 'none';
```
seanmhoffman: I noticed when implementing the sortable component inside my app that after the initial drag and reordering of items `freeze` is invoked on every item causing a `transition: none 0s ease 0s` style to be applied. Unfortunately it doesn't look like this is cleaned up anywhere and causes the items to stop transitioning. Was this intended?

nlfurniss: Are you saying your issue is due to the changes in this PR or that this issue is there no matter what?

Either way it would be much better to create an issue. A comment on a closed PR isn't going to get a lot of attention

seanmhoffman: For me it is an issue because its stops some transitions I have in place when dragging occurs. I didn't see anywhere stated in the docs that describes that situation. However I'm not sure if its an actual issue with the addon or if it was intended. That's why I posed the question here instead of creating an issue for it, I didn't want to create an issue for something thats a feature and not a bug.

jgwhite: We didn’t intend to prevent people applying transitions when dragging occurs, but looking back it seems like this fell into the “unspecified behaviour” category. The history of this `freeze` method dates back to [some fixes I added when Glimmer 1 first arrived](https://github.com/heroku/ember-sortable/pull/21) nearly 4 years ago. I don’t remember might thought processes back then but it’s totally possible there’s a better way to solve this that doesn’t involve stripping *all* transitions.

PR #216, opened 2018-12-10
Fix incorrect ScrollContainer dimension calcs

nlfurniss:
My last PR incorrectly was calling `getComputedStyle` on `window` doesn't work, so current master will throw an error.

Used the same calc that jQuery uses to do `$(window).width()/height()`


jgwhite:
Getting some odd build errors on Travis unrelated to the change. I’m gonna try bumping ember-try.

jgwhite:
Retrying the build again — sorry for letting this sit for so long!

PR #221, opened 2018-12-18
Pass event to _preventClickHandler

fonkgoku:
See: https://github.com/heroku/ember-sortable/issues/218


fonkgoku:
Sure thing :), thanks for the release! Will update right away ;)

PR #223, opened 2018-12-19
fixes for Firefox

fivetanley:
- SVGs don't have height in Firefox, so we have to fallback to the
parentElement's height in order to calculate scale correctly. When the
`drag` handler was used on a `.handle` element that was an SVG, the
value of `scale` was always zero. Now it is the correct value (usually
1).
- Use the `transitionend` event to wait for events to finish. This prevents race conditions versus trying to extract the time from the styles. In chrome,
these animations were scheduled in such a way that the dom elements were
technically in the right order but visually inaccurate if we slowed the
transitions down. In Firefox, the smoke tests failed due to the elements
being out of order. In addition to using the `transitionend` event
(which is supported by all browsers we
support(https://developer.mozilla.org/en-US/docs/Web/Events/transitionend#Browser_compatibility),
when `DEBUG` mode is active (e.g. development or test builds, not
production builds`, we emit a custom event on the document and use a
custom waiter to prevent Ember from advancing the test suite too fast
before the transitions are done (the transforms need to finish so that
the elements are in the correct order in the dom).
- Add Firefox to our .travis.yml and testem.js files so that we test
against Firefox in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ sudo: false
 dist: trusty
 
 addons:
+  firefox: latest
   chrome: stable
 
 cache:
--- a/addon/helpers/drag.js
+++ b/addon/helpers/drag.js
@@ -54,11 +54,18 @@ export function drag(app, mode, itemSelector, offsetFn, callbacks = {}) {
     let offset = offsetFn();
     let itemElement = item.get(0);
     let rect = itemElement.getBoundingClientRect();
-    let scale = itemElement.clientHeight / (rect.bottom - rect.top);
+    // firefox gives some elements, like <svg>, a clientHeight of 0.
+    // we can try to grab it off the parent instead to have a better
+    // guess at what the scale is.
+    // https://bugzilla.mozilla.org/show_bug.cgi?id=874811#c14
+    // https://stackoverflow.com/a/13647345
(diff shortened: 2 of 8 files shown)


jgwhite:
Released in 1.12.4

BnitoBzh:
Breaks all drag & drop usages on my applications .. (tested on Safari, Chrome and Firefox) ...
All worked fine before 1.12.4 .... 
Now after dragging an element, the onUpdate action is not called and the dragged element is not draggable anymore ...

jgwhite:
@BnitoBzh very sorry to hear that. Are you able to give us any code snippets that will help us figure out what happened here? Are you OK to downgrade for the time being?

BnitoBzh:
@jgwhite I have no code snippet, i am using ember-sortable in its basic way, (readme example).
I have downgraded and blocked ember-sortable to its version 1.12.3.
I have not a lot of time now to debug this issue, sorry

jgwhite:
@BnitoBzh no problem. We’ll look into it and make a note on this thread when a fix lands.

fivetanley:
@BnitoBzh Can you open this as a new issue? Thanks

apellerano-pw:
Am having the same bug where `transitionend` never fires, I'll open an issue with what I know

PR #224, opened 2019-01-02
[bugfix] default `dx` and `dy` to 0 for drag helper

fivetanley:
When using `drag` in tests, it's common to only drag vertically or
horizontaly, and leave either `dx` or `dy` undefined. When `dx` or `dy`
is undefined, the math calculated by the `drag` helper returns NaN
(because multiplying by undefined or null returns `NaN` instead), which
can throw errors when performing the drag in tests.

This fixes the helper by defaulting `dx` or `dy` to 0, a valid number,
instead.


jgwhite:
Released in 1.12.4

Issue #226, opened 2019-01-03
Can't use `drag` test helper when using a handle


kdamball:
Without using a `handle` prop for `{{#sortable-item}}`, one can easily use the `drag` test helper. However, without a `handle` prop - you can't use inputs as per https://github.com/heroku/ember-sortable/issues/124. In other words, impossible to run acceptance tests in this scenario.

kdamball:
I managed to reproduce it here: https://ember-twiddle.com/78875ceb8f9ea82e63026ec71727d6d6?openFiles=tests.acceptance.my-acceptance-test.js%2C

jgwhite:
Thanks for the report and the awesome reproduction.

There seem to be a couple of things at work:

- The way we calculate `scale` in the drag helper (recently fixed)
- The way we determine whether or not animated transitions are in play (still an open question)

Could you try the latest release (1.12.4)?

I was able to get the twiddle passing by bumping ember-sortable to 1.12.4 and adding the following CSS:

```css
.sortable-item {
  transition: none;
}
```

The fact that an unspecified transition leads to an incomplete sort is a larger problem, but one we can tackle separately.

kdamball:
CSS transitions leading to elements not being sorted is probably the bigger issue here 😅

PR #227, opened 2019-01-04
Tweak clientHeight calculation in drag helper

jgwhite:
@fivetanley while investigating #226 I noticed this discrepancy.


mixonic:
I presume this is blocked on a test- seems funky that it wasn't caught by anything.

jgwhite:
> seems funky that it wasn't caught by anything

True. I would have thought the Firefox side of CI would have caught this.

jgwhite:
@fivetanley could I get your review on this?

PR #229, opened 2019-02-11
inline `touch-action:none` not being properly applied

froskos:
When using `document.querySelector(this.get('handle'))` in order to find the element with the dragging handle, the first matching element in the whole DOM is returned. 

The consequence is that only the first element in a sortable list has the inline style `touch-action:none`. If the list had *n* members, the current code would match the first element in the list *n* times, applying the style only to that element *n* times in an idempotent way.

The fix scopes the query to the actual element which has just been inserted in the DOM so as to guarantee that each draggable element gets the inline style.

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -255,7 +255,7 @@ export default Mixin.create({
 
     // Instead of using `event.preventDefault()` in the 'primeDrag' event,
     // (doesn't work in Chrome 56), we set touch-action: none as a workaround.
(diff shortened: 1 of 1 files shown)


froskos:
This PR is meant to solve issue #230 :

https://github.com/heroku/ember-sortable/issues/230

jgwhite:
Released as part of [1.12.5](https://github.com/heroku/ember-sortable/releases/tag/1.12.5). Thanks again!

PR #236, opened 2019-02-16
getX and getY methods do provide now for non-wrapped events too (fixe…

froskos:
…s problem of silent error at _prepareDrag method happening in touch devices )


jgwhite:
Thanks for the PR! This seems solid to me. One of ember-try variants failed with a segfault so I’ve restarted it.

jgwhite:
Released in [1.12.6](https://github.com/heroku/ember-sortable/releases/tag/1.12.6)

PR #242, opened 2019-04-23
Runloop tweaks

mixonic:
Fixes https://github.com/heroku/ember-sortable/issues/219

I've started at the tests a bit and I'm not sure how to test this. The problem is that to test it I need test helpers that don't add a runloop themselves (which is an incorrect many test helpers used to do).

I'll first need to ascertain that the test helpers really are adding runloop wrapping.

### DIFF
--- a/addon/mixins/sortable-item.js
+++ b/addon/mixins/sortable-item.js
@@ -395,7 +395,9 @@ export default Mixin.create({
       dragActions.forEach(event => window.removeEventListener(event, dragThrottled));
       endActions.forEach(event => window.removeEventListener(event, drop));
 
-      this._drop();
+      run(() => {
+        this._drop();
+      });
     };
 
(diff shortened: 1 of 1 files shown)


mixonic:
Here is a drag event being triggered in test, and you can see it is wrapped in a runloop by the testing harness.

I'm personally hoping to not break support back to 1.12 in the short term 😬 and I'm not sure of the implications but I think https://github.com/heroku/ember-sortable/pull/211 may address this by avoiding the `andThen` system in favor of async/await.

jgwhite:
Thanks for this. Sorry to be a pain but could you rebase this atop master? I had to do a bit of housekeeping.

jgwhite:
Scratch that last message — This change seems worth landing so I went ahead and rebased.

PR #243, opened 2019-04-24
1.12.7

jgwhite:


jgwhite:
I might be best to land #227 and #244 first, and bundle them into this release.

PR #245, opened 2019-04-24
Add Ember 2.8, 2.12 to testing. Test Node 8.

mixonic:
I'd like to ensure ember-sortable remains backwards compatible to Ember 1.12. I'd welcome any feedback from maintainers about if that is something they are OK with, or if not then to get some reasoning about what the base version should be.

Regardless this PR is an incremental change to add Ember 2.8 and 2.12 support to the test suite. I'll follow up on any failures.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "6"
+  - "8"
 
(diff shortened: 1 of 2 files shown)


mixonic:
Removing the WIP, this passes without further changes.

jgwhite:
Speaking just for myself, I see no reason not to commit to maintaining compatibility back to 1.12. If the commitment blocks us from making progress in some future dimension that maybe we revisit the constraint.

PR #249, opened 2019-04-30
move @ember/jquery to devDependency

### LINKED ISSUE #248: Make @ember/jquery a developer dependency
Forgive me if I'm missing something but since jQuery was replaced with native javascript as part of #212, why do we still list `@ember/jquery` as a package dependency?

I note that it is still used for tests but wouldn't that just require it to be a dev dependency?

fran-worley:
Fixes #248

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,11 +22,11 @@
     "test:all": "ember try:each"
   },
   "dependencies": {
-    "@ember/jquery": "^0.6.0",
     "ember-cli-babel": "^7.1.2",
     "ember-cli-htmlbars": "^3.0.0"
   },
   "devDependencies": {
+    "@ember/jquery": "^0.6.0",
     "@ember/optional-features": "^0.6.3",
     "broccoli-asset-rev": "^2.7.0",
     "ember-cli": "~3.7.1",


jgwhite:
I am 👍 but before we ship this I’d like to try to find an app that uses the helpers and see what happens. I suspect it’ll be fine as the Ember helpers are still public API.

jgwhite:
Sorry for the delay.

I did a bit more research and here’s what I found:

If a consuming app [has jQuery disabled](https://guides.emberjs.com/release/configuring-ember/optional-features/#toc_jquery-integration) then all “classic” test helpers (including ember-sortable’s `drag` and `reorder`) will emit the following error:

```
If jQuery is disabled, please import and use helpers from @ember/test-helpers.
```

This is good news for this PR, as it means that moving `@ember/jquery` to devDependencies won’t cause any regressions. Anyone using `drag` or `reorder` with jQuery disabled will already see the above message in the current release.

Longer term, ember-sortable should provide equivalent test helpers in the new style (i.e. `import { drag, reorder } from 'ember-sortable/test-helpers'`) but that’s a matter for a future PR.

jgwhite:
Thanks so much for the contribution!

jgwhite:
Released in 1.12.8

PR #251, opened 2019-05-06
Add warning to polyfill `Element.closest()`

### LINKED ISSUE #241: Element.closest() is unsupported in IE
**Describe the bug**
The use of `Element.closest()` was introduced in #212, causing the `_primeDrag` event to fail on IE 10/11 as the method is unsupported.

**To Reproduce**
1. Open the demo pages in IE10
2. Observe JS errors

**Additional context**
I think this just needs a doc change to call out browser support explicitly, and warning that consumers will need to polyfill in app code.

Happy to raise a PR for this!

samhogg:
Fixes #241

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ Version 1.0 depends upon the availability of 2D CSS transforms.
 Check [the matrix on caniuse.com](http://caniuse.com/#feat=transforms2d)
 to see if your target browsers are compatible.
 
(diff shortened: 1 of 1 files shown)


btecu:
Adding a warning doesn't really fix the issue. The addon should avoid using `Element.closest`.

samhogg:
@btecu open to that, too - although this was what I proposed in #241.

I don't think it's unreasonable to ask consumers of the library to polyfill, but can see the benefits of avoiding `Element.closest()` altogether.

jgwhite:
While it’s true that this isn’t a solution, it’s definitely an improvement. I’ll reopen the issue once this is merged.

jgwhite:
Thanks @samhogg!

PR #257, opened 2019-06-04
Created importable modern test helpers

ygongdev:
Currently for reference.

### DIFF
--- a/addon/helpers/native.js
+++ b/addon/helpers/native.js
@@ -1,0 +1,15 @@
+/**
+ * Vanilla Javascript equivalent of JQuery's `.offset`.
+ * Reference: https://github.com/nefe/You-Dont-Need-jQuery
+ *
+ * @method getOffset
+ * @param {Element} el an element
+ */
+export function getOffset(el) {
+  const box = el.getBoundingClientRect();
+
+  return {
+    top: box.top + window.pageYOffset - document.documentElement.clientTop,
+    left: box.left + window.pageXOffset - document.documentElement.clientLeft,
+  };
+}
--- a/addon/helpers/reorder.js
+++ b/addon/helpers/reorder.js
@@ -1,4 +1,7 @@
 import { registerAsyncHelper } from '@ember/test';
+import { settled, findAll } from '@ember/test-helpers';
+import { drag } from '../helpers/drag';
+import { getOffset } from '../helpers/native';
 
 /**
   In tests, the dummy app is rendered at half size.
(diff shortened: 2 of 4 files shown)


addon/helpers/native.js
```diff
@@ -0,0 +1,15 @@
+/**
```
nlfurniss: Unless this file is used by the app (outside of tests), it should live in addon-test-support

ygongdev: Interestingly, `drag` and `reorder` are also not used by the app.
`waiter` also seems to be used only by the test, but is wrapped inside a `DEBUG` flag in the app.
```
document.dispatchEvent(new Event('ember-sortable-drop-stop')
```
Sounds like they all can be moved into `addon-test-support`?

ygongdev:
If someone could give some feedback or knows the answer to these questions that'd be great
1. I couldn’t find a suitable vanilla equivalent for jquery’s `filter`. Looks like it can take jquery functions? e.g `:contains`. That’s why I commented out some tests that was using `:contains`.
2. `Acceptance smoke` tests are failing horribly. Could be a combination of me migrating the test helpers incorrectly and test setup. I'm not exactly an expert at migrating the old syntax to the new one.

addon/helpers/native.js
nlfurniss: I would say yes except that would require a major version bump and complicates your PR. Prob best left for another time

nlfurniss:
Why can't you use `filter`?

Can you post a gist of the test failures? Hard to know where to begin without any failures to look at

ygongdev:
@nlfurniss native `Array.filter` doesn't account for jquery extensions that are not part of the CSS Specification.
e.g, `Array.filter(:contains)` will throw an error because `:contains` is a jquery selector.

addon/helpers/reorder.js
```diff
@@ -30,11 +33,49 @@ const OVERSHOOT = 2;
   @return {Promise}
 */
 
-export function reorder(app, mode, itemSelector, ...resultSelectors) {
+export async function reorder(mode, itemSelector, ...resultSelectors) {
+  await settled();
+
+  resultSelectors.forEach(async (selector, targetIndex) => {
```
gmurphey: I believe we can't use `Array.prototype.forEach` here since it can't `await` async operations. https://github.com/babel/babel/issues/909

ygongdev:
Closing this in favor of #271

PR #261, opened 2019-06-15
eslint: enforce space before and after arrow of arrow function

stephencattaneo:
Have eslint enforce that we want `() => {}` not `()=>{}`


jgwhite:
Thanks!

PR #265, opened 2019-07-09
Properly convert ScrollContainer width and height to integers

benedikt:
This fixes the broken scrolling for elements inside overflow: auto/scroll containers.

### DIFF
--- a/addon/system/scroll-container.js
+++ b/addon/system/scroll-container.js
@@ -13,8 +13,8 @@ export default class ScrollContainer {
       let { top, left } = this.element.getBoundingClientRect();
       this.top = top;
       this.left = left;
-      this.width = getComputedStyle(this.element).width;
-      this.height = getComputedStyle(this.element).height;
+      this.width = parseFloat(getComputedStyle(this.element).width);
+      this.height = parseFloat(getComputedStyle(this.element).height);
       this.scrollWidth = element.scrollWidth;
       this.scrollHeight = element.scrollHeight;
     }


ygongdev:
@benedikt Hi I would love to merge this, but I want to get some context on the issue.
Do you have a demo of the bug available by any chance?
I want to see if we're ok with converting it to `int` because I think the box model actually preserves the `decimals` values and use it to calculate layout

benedikt:
@ygongdev I don't have a demo up, but the issue boils down to this: `.width` and `.height` on the results of `getComputedStyle` return a string including a `px` suffix. As a result, any calculations afterwards return `NaN`.

ygongdev:
@benedikt got it got it. That  makes a lot of sense. What do you think about of using `parseFloat` instead to preserve the true value as much as possible?

benedikt:
@ygongdev Yes, that should work as well. I'll update the pull request accordingly.

benedikt:
@ygongdev I just pushed the `parseFloat` variant. 

FYI: There are a couple more `parseInt` calls in other places as well. It probably makes sense to update those eventually.

addon/system/scroll-container.js
```diff
@@ -13,8 +13,8 @@ export default class ScrollContainer {
       let { top, left } = this.element.getBoundingClientRect();
       this.top = top;
       this.left = left;
-      this.width = getComputedStyle(this.element).width;
-      this.height = getComputedStyle(this.element).height;
+      this.width = parseFloat(getComputedStyle(this.element).width, 10);
+      this.height = parseFloat(getComputedStyle(this.element).height, 10);
```
ygongdev: `parseFloat` doesn't have a 2nd param. It doesn't need a radix unlike `parseInt`.

ygongdev:
@benedikt if you want feel free to update those as well! I left a comment.

addon/system/scroll-container.js
benedikt: Done ✅

PR #266, opened 2019-08-21
update ember-cli to 3.12

fivetanley:
This also changes `ember-try` to no longer test against Ember 2.x since it is no longer LTS under Ember's release cycle and not receiving security updates. The earliest LTS we now supports is 3.4.

In this PR I also used `ember-cli-update --run-codemods` to update our tests to the latest and greatest test syntax.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,7 @@
 module.exports = {
   root: true,
   parserOptions: {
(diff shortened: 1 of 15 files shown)


ygongdev:
Closing as `ember-cli` was bumped to `3.14`

PR #270, opened 2019-09-08
Removing sortable-item-mixin

ygongdev:
1. Removing sortable-item-mixin. This will get ember-sortable one step closer to native classes.
    2. Combined sortable-item tests with sortable-item-mixin tests. 
    3. Moved functions outside of component into utils.

I think the mixin is safe to remove, I don't think anyone would be using the mixin.

Edit: Opps idk why it pulled in the other commits.
Edit2: Fixed it


nlfurniss:
Nice!

ygongdev:
@jgwhite I think this shouldn't introduce any backward incompatible and could be merged? It'll help us toward upgrading to Ember 3.x.x

PR #271, opened 2019-09-08
Modernizing test suite

ygongdev:
As an effort to modernize the test suite, I
- [x] Added `data-test-selector` library and used it in the test's `application.hbs`
- [x] Moved `drag` and `reorder` test-helpers into `test-support`.
- [x] Upgraded `drag` and `reorder` to the new syntax.
- [x] Removed `registerAsyncHelper`.
- [x] Converted the `Acceptance` test to the newest syntax.
- [x] Removed all JQuery usage inside the tests.
- [x] ~~Fix the new `Acceptance` test and/or the app code. Right now, the `Accpetance` tests are super flaky, From my observations, in `reorder`, the DOM seems to be in the wrong order on subsequent iterations within `reorder`. E.g When I queried the given selector, I found that the last element of the array pointed to the first element in the DOM, etc. 
The reason why the old tests passed is because of we used `:contains` to grab the element with the corresponding text, so we will still grab the correct element despite its position might not be where we expect it to be. However, I suppose we can change the definition of `reorder` to accept a function that does something similar to `:contains`, but I think the `reorder` helper then becomes more complicated and not worth using.~~
Tests are mostly stable now.
 
~~Feel free to fix the flakiness of the tests. Once that is fixed, I think the PR is will be good to go 😃~~ 
This would require a major bump. FYI, this is an improved version of #257 @jgwhite @nlfurniss 
Related issues: #272


test-support/ember-sortable/helpers/reorder.js
```diff
@@ -0,0 +1,51 @@
...
+    const sourceElement = find(resultSelectors[targetIndex]);
+    const targetElement = items[targetIndex];
+    const dx = getOffset(targetElement).left - OVERSHOOT - getOffset(sourceElement).left;
+    const dy = getOffset(targetElement).top - OVERSHOOT - getOffset(sourceElement).top;
+
+    await drag(mode, sourceElement, () => { return { dx: dx, dy: dy }; });
+  }
+}
+
+function getOffset(el) {
```
lifeart: can we extract getOffset to test-support/ember-sortable/utils/offset ?
same method declared in two files, we can reuse one declaration

ygongdev: Yeah sure. Nice catch.

Issue #273, opened 2019-09-20
1.x to 2.x Migration RFC


ygongdev:
[Rendered](https://github.com/ygongdev/ember-sortable/blob/migration-doc/V2_MIGRATION_RFC.md)
Please review the RFC. Any comments  would be appreciated.
@jgwhite @H1D @chriskrycho @drewlee

chriskrycho:
Thanks for writing it up! For reference, so folks can easily comment inline on it, here's a link to the branch compared against master (which I'll update later if need be): https://github.com/ygongdev/ember-sortable/commit/migration-doc

ygongdev:
Good callout haha.
I created a PR, so we can

jgwhite:
Thanks so much for writing this up. It’s awesome.

I’d encourage you to add a rough release plan to this RFC. Trying to introduce all the features mentioned in the RFC (alongside a good migration story) all in one release sounds too ambitious to me.

i.e.

```md
## 2.0

- New API with backward incompatible changes
- Migration guide

## 2.1

- Built-in keyboard support

## 2.2

- Support for nesting
```

ygongdev:
@jgwhite 

Good idea,  I'll try to break it down into multiple version changes.

dbashford:
Quick question on this... is it possible to get `volatile`/`sendAction` fixes without 2.0 and breaking changes? That in the cards?

ygongdev:
@dbashford I don't have much context on `volatile`, but if the solution is migrate it into native classes, we would also need to bump the `Ember` version to `3.x.x`, which might introduce some other breaking changes, so might as well do it all in one.

Also for `sendAction`, I think that's very do-able, either convert to `this.{actionName}` or use `tryInvoke` if the action might not exist.

@jgwhite @chriskrycho @H1D 
I just added a `release plan` into the `RFC`. If it looks good to you, I think we can get started on 
**2.0**. Although should we  decide on which Ember.version we would to go with?

chriskrycho:
I’ll be reviewing the RFC in detail tomorrow! More thoughts then!

ygongdev:
FYI, I created a new branch `latest-1.x.x` to record the latest backwards compatible master. This is done so we can proceed to active develop on `master` with backwards incompatible changes.

PR #276, opened 2019-09-23
Avoid using `volatile` computed properties

alexlafroscia:
Using `.volatile` has been deprecated as of Ember 3.9. A native gettter
or setter should be used instead.

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -39,22 +39,22 @@ export default Component.extend({
     @property itemPosition
     @type Number
   */
-  itemPosition: computed(function() {
+  get itemPosition() {
     let direction = this.get('direction');
 
     return this.get(`sortedItems.firstObject.${direction}`) - this.get('sortedItems.firstObject.spacing');
-  }).volatile(),
+  },
 
   /**
     @property sortedItems
     @type Array
   */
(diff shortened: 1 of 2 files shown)


alexlafroscia:
Everything seems to break when refactoring things this way -- the context inside the getters is not the component instance, for some reason. Need to look into the right way to approach this without moving to using ES6 classes entirely.

jgwhite:
Thanks for the PR.

There’s a small group of folks currently looking into how to transition to a v2 and potentially unlock changes like this. Take a look at https://github.com/heroku/ember-sortable/issues/273 for more.

This seems like a good candidate to be addressed in a v2.

jgwhite:
> Maybe ember-compatibility-helpers can be used to use .volatile in the older versions where it is not deprecated and a getter in versions that is it supported.

That sounds like a good approach.

ygongdev:
@alexlafroscia 
Heh, this could totally be a discussion point for the RFC in #273 
My take is that we want to adopt as many new cool stable Ember features as we can, but also make sure we don't leave the vast majority of users, whose apps might be incompatible, behind.

ygongdev:
Should be completed in #322

ygongdev:
@dbashford wanna give `2.0.3` a shot? 😉

PR #279, opened 2019-10-03
Closest polyfill

ygongdev:
1. Adding element-closest package to support IE9+ support for Element.closest
    2. Adding note to README regarding which version will polyfill it.

Solves #241 
More than glad to have a better way to polyfill this. Since I only see it used in `sortable-item`, I only polyfilled there...

### DIFF
--- a/addon/utils/polyfill/closest.js
+++ b/addon/utils/polyfill/closest.js
@@ -1,0 +1,43 @@
+// Polyfill for 'closest' and `matches` in IE11.
+// https://github.com/jonathantneal/closest
+export default function polyfill(window) {
+  if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+    (function(ElementProto) {
+      if (typeof ElementProto.matches !== 'function') {
+        ElementProto.matches =
+          ElementProto.msMatchesSelector ||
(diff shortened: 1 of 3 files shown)


addon/utils/polyfill/closest.js
```diff
@@ -0,0 +1,43 @@
+// Polyfill for 'closest' and `matches` in IE11.
+// https://github.com/jonathantneal/closest
+export default function polyfill(window) {
+  if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+    (function(ElementProto) {
+      if (typeof ElementProto.matches !== 'function') {
```
btecu: Here's what I've been using for IE11:
```
if (!Element.prototype.closest) {
    Element.prototype.closest = function(selector) {
      let element = this;

      do {
        if (element.msMatchesSelector(selector)) {
          return element;
        }

        element = element.parentElement || element.parentNode;
      } while (element !== null && element.nodeType === 1);

      return null;
    };
  }
```

Note that Ember only supports IE11, so you don't need to support older IE browsers.

knownasilya:
You could also do async import based on need.

ygongdev:
Closing in favor of #280 , which polyfills on the app level.

PR #280, opened 2019-10-05
Closest polyfill

ygongdev:
1.  Adding closest polyfill
2. Adding downloads badge

Manually tested on IE11.
Addresses #241

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # Ember-sortable
 
 [![npm version](https://badge.fury.io/js/ember-sortable.svg)](http://badge.fury.io/js/ember-sortable)
+[![Download Total](https://img.shields.io/npm/dt/ember-sortable)](http://badge.fury.io/js/ember-sortable)
 [![Ember Observer Score](http://emberobserver.com/badges/ember-sortable.svg)](http://emberobserver.com/addons/ember-sortable)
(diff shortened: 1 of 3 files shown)


btecu:
@ygongdev you didn't have to polyfill `matches`, just `closes`:
```
  if (!Element.prototype.closest) {
    Element.prototype.closest = function(selector) {
      let element = this;

      do {
        if (element.msMatchesSelector(selector)) {
          return element;
        }

        element = element.parentElement || element.parentNode;
      } while (element !== null && element.nodeType === 1);

      return null;
    };
  }
```

ygongdev:
Oh ok, I will remove the matches polyfill

btecu:
You have to update the `closest` as well, to use `msMatchesSelector`  - `(element.msMatchesSelector(selector)`, since we're literally putting this polyfill in place for IE.

ygongdev:
@btecu We don't need to make specific distinction right? As in, `element.matches` should be fine, since we're targeting `closest`, not `matches`? Doing so would make other non IE browsers use `msMatchesSelector`

btecu:
The non IE browsers already implement `matches` so `if (!Element.prototype.closest) {` prevents us from overriding their `matches` implementation.

PR #252, opened 2019-05-14
use only closure actions internally

H1D:
Another simple PR to get familiarized with codebase.

This PR changes all `this.sendAction(XXX)` to direct invocation `this.XXX()`. Fixes #213 

Notes:

* ember > 1.13 needed (not an issue since we require Ember >=2.4 for `ember-sortable 2.x` )
* `sendAction()` was [deprecated](https://deprecations.emberjs.com/v3.x/#toc_ember-component-send-action) in 3.4 and will be removed in 4.0

PR #282, opened 2019-10-05
Deprecating sendAction

ygongdev:
This will cause backward incompatbility if a consumer is passing in `actions` as `strings`, e.g
`onDragStart="onDragStart"` instead of `onDragStart=(action "onDragStart")`


ygongdev:
References #252. #252 also contains some contextual component changes, which I plan to push as a separate commit.

PR #290, opened 2019-10-18
Adding built-in keyboard support and focus management

ygongdev:
1. Added keyboard support
2. Added focus manage

TODOs:
- Fix tests
- Work on migration doc (I think we might deprecate some old APIs to encourage people to use better and accessible approaches, which this new addition is meant to tackel).

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -3,6 +3,15 @@ import Component from '@ember/component';
 import { set, get, computed } from '@ember/object';
 import { run } from '@ember/runloop';
 import layout from '../templates/components/sortable-group';
+import {
+  isEnterKey,
+  isSpaceKey,
+  isEscapeKey,
+  isDownArrowKey,
+  isUpArrowKey,
+  isLeftArrowKey,
+  isRightArrowKey,
+} from '../utils/keyboard';
 
 const a = A;
 const NO_MODEL = {};
@@ -10,7 +19,7 @@ const NO_MODEL = {};
 export default Component.extend({
   layout,
 
-  attributeBindings: ['data-test-selector'],
+  attributeBindings: ['data-test-selector', 'tabindex', 'role'],
 
   /**
     Called when order of items has been changed
@@ -38,6 +47,24 @@ export default Component.extend({
   model: NO_MODEL,
 
   /**
+   * @property a group associated with the model
+   * @type Any
+   * @default null
+   */
+  groupModel: NO_MODEL,
+
+  /** Primary keyboard utils */
+  // Tracks the currently selected item
+  selectedItem: null,
+  // Tracks the current move
+  move: null,
+  // Tracks the status of keyboard reorder mode
+  isKeyboardReorderModeEnabled: false,
+  // Tracks if we're still performing a programmatical focus.
+  isRetainingFocus: false,
+  /** End of keyboard utils */
+
+  /**
     @property items
     @type Ember.NativeArray
   */
@@ -50,21 +77,28 @@ export default Component.extend({
     @type Number
   */
   itemPosition: computed(function() {
-    let direction = this.get('direction');
+    const direction = this.get('direction');
 
     return this.get(`sortedItems.firstObject.${direction}`) - this.get('sortedItems.firstObject.spacing');
   }).volatile(),
 
   /**
+   * An array of DOM elements.
     @property sortedItems
     @type Array
   */
   sortedItems: computed(function() {
-    let items = a(this.get('items'));
-    let direction = this.get('direction');
+    const items = a(this.get('items'));
+    const direction = this.get('direction');
 
     return a(items.sortBy(direction));
   }).volatile(),
+
+  init() {
+    this._super(...arguments);
+
+    this.set('moves', []);
+  },
 
   /**
     Register an item with this group.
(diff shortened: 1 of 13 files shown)


addon/components/sortable-group.js
```diff
@@ -3,14 +3,24 @@ import Component from '@ember/component';
 import { set, get, computed } from '@ember/object';
 import { run } from '@ember/runloop';
 import layout from '../templates/components/sortable-group';
+import { mutateDOM, readDOM } from 'ember-batcher';
```
lifeart: do we really need `ember-batcher`? (it can cause side-effects if event will-be re-fired)

.huskyrc
```diff
@@ -1,6 +1,6 @@
 {
   "hooks": {
-    "pre-commit": "yarn lint",
-    "pre-push": "yarn lint && yarn test"
+    "pre-commit": "yarn lint && yarn test",
```
lifeart: I think it's out of the scope of current PR

addon/components/sortable-group.js
```diff
@@ -163,10 +187,202 @@ export default Component.extend({
...
+      readDOM(() => {
+        this.element.focus();
+        this.set('isRetainingFocus', false);
+      });
+      // Prevent the default scroll
+      event.preventDefault();
+      return;
+    }
+
+    if (isKeyboardReorderModeEnabled) {
```
lifeart: can we extract it into new method? like `handleKeybordReorder`

addon/components/sortable-group.js
```diff
@@ -163,10 +187,202 @@ export default Component.extend({
...
+            handle.focus();
+          } else {
+            // The consumer did not use a handle, so we set focus back to the item.
+            itemElement.focus();
+          }
+        });
+      } else if (isEscapeKey(event)) {
+        this.cancelKeyboardSelection();
+        this.set('isRetainingFocus', true);
+        readDOM(() => {
```
lifeart: it will be nice to call it as different method as well, with ability to cancel it if we reenter here.

addon/components/sortable-group.js
ygongdev: Without `ember-batcher`, several things wouldn't work properly,
For example,
1. `focus` does not work after we "move" an item. My hypothesis is that it might be scheduled in an earlier DOM operation, so `ember-batcher` allows me to batch the `focus` with the correct DOM update(where the items are swapped). So far it works as intended.
2. Rapid re-firing of an `arrow` event actually causes UI render race conditions. Using `ember-batcher` for `moveItem` actually solved this issue. My understanding is that even if we fire off multiple duplicate events (multiple `mutateDOM` calls), all of them will still be batched for the next DOM update.

Can you elaborate on `side-effects of `ember-batcher`. 
If we are worried about `side-effects` for re-firing, we can always look at other alternatives.

.huskyrc
ygongdev: I just switched the `test` to the commit hook, so I can `-n` my commit and not run the test when I push b/c I know the tests are going to fail for now haha

addon/components/sortable-group.js
ygongdev: Yup, will do that in the final PR. I got lazy and just wanted to get something working without spending too much time on making the code nice

addon/components/sortable-group.js
ygongdev: 👍

ygongdev:
1. Added `a11yAudit` and fixed several color violations.
2. Added keyboard tests

I had to duplicate my `keyboard` helpers because I couldn't make it discoverable during `testing`, any recommendation to get rid of this would be appreciated.

addon/components/sortable-group.js
ygongdev: I was discussing with some folks and they recommended the `ember-batcher` over `scheduling`. Looks like `scheduleOnce(afterRender` might trigger more unnecessary forced layouts?
However, I did notice that `ember-batccher` is only supported for `ember 3.4`^ so to keep support for `ember-2.8` -> `ember-3.4` maybe I'll need to leverage something else.

ygongdev: using `scheduleOnce('render')` for now.

addon/components/sortable-group.js
ygongdev: I decided to abstract the `focus` logic, but not the entire logic because it seemed a bit unnecessary imo. we have `cancelKeyboard` and then focus logic, which should be readable.

addon/components/sortable-group.js
```diff
@@ -136,17 +162,140 @@ export default Component.extend({
     @method commit
   */
   commit() {
-    let items = this.get('sortedItems');
-    let groupModel = this.get('model');
-    let itemModels = items.mapBy('model');
-    let draggedItem = items.findBy('wasDropped', true);
+    const items = this.get('sortedItems');
```
kategengler: This seems to remove the concept of a `groupModel` (https://github.com/adopted-ember-addons/ember-sortable#declaring-a-group-model) and is a breaking change. Is that intentional?

ygongdev: Yes I intentionally strip that out. The new 2.0.0 release will contain a lot of breaking changes (`master` already contains many), so this is going to be a major part of it.

The sole purpose of the `groupModel` was so that you have a reference to the `group` to set the new sorted array to, which I consider unnecessary as the consumer has access to the `group` without feeding it to `ember-sortable`.
```
onChange(models, draggedModel) {
 this.group.set('items', models);
}
```

My intention is to remove the bookkeeping overhead and keep the addon simple. which should be `you supply us with an array of items you want to arrange and we will give you back the new sorted order`. The added benefit of `group` seems to be trivial. However, I'm open to feedback on why maybe we should keep it.

ygongdev:
FYI screen reader announcements and visual indicators will be tackled next.

addon/components/sortable-group.js
kategengler: It's definitely a breaking change; I haven't used this addon in quite awhile, so I don't if it is a big deal or not. That said, it seems like this PR should make some changes to the README and other documentation.

ygongdev: I plan on update all the README and documentation when all the preparation for the next release is ready. We're not planning to release anything until the first major release is ready.

ygongdev: Imo, I think changing the README now might confuse people to think that the current release (which is way behind) contain these changes when they come and look at the repo.

ygongdev: I plan to create a `migration doc`, update `readme` and other documentations when all the implementations are completed prior to the next release. Since right now we might still need to re-iterate and have the overhead of keep updating the docs.

ygongdev: @kategengler, actually I added the `groupModel` support...It might be useful for future feature requests that I'm getting. But regarding README and other documentation, are you ok with me holding off on that until after the migration to v2 is done?

kategengler: 👋 I'm not a maintainer of this repo, was just making a suggestion, earlier 😀

ygongdev: Ah ok, appreciate your feedback!

Issue #304, opened 2019-10-31
2.0.0 Could not find module ember-test-waiters imported from ember-sortable/components/sortable-item


aandis:
**Describe the bug**
After upgrading to `2.0.0` you see the error -

> Could not find module `ember-test-waiters` imported from `ember-sortable/components/sortable-item`

**To Reproduce**
Upgrade to `2.0.0`

**Expected behavior**
Addon should work normally.

**Screenshots**
![error](https://user-images.githubusercontent.com/4983347/67972348-80263f80-fc34-11e9-82e3-e7d2a6477d28.png)

ygongdev:
@aandis 🤔 
Thanks for the report, I'll look into it.
I think it's because `ember-test-waiters` was only added in `devDependencies`.

ygongdev:
Does this prevent you from using the addon entirely? I'll release a patch soon

aandis:
> Does this prevent you from using the addon entirely?

Yeah. Thanks for looking into it so quick.

ygongdev:
Ok, I just released `2.0.1`, let me know if that solves your issue.
Please report any more bugs that you find. Thanks!

gtb104:
For me, 2.0.1 fixed this import error.

PR #305, opened 2019-10-31
[Bugfix] ember-test-waiters

ygongdev:
Moving ember-test-waiters and ember-test-selectors from devDependencies to dependencies to solve #304


gtb104:
Nice! I was just about to point this out.

Issue #306, opened 2019-11-04
Uncaught TypeError after 2.0.1 upgrade


gtb104:
**Describe the bug**
Getting Uncaught TypeError after upgrading to 2.0.1.

**To Reproduce**
1. Going to a page that has an ember-sortable component causes an Uncaught TypeError, and the page fails to load.

**Expected behavior**
No errors

**Screenshots**
![image](https://user-images.githubusercontent.com/1440968/68126294-7c6b1500-fee1-11e9-972b-8e77b0707e4b.png)

**Additional context**
This is my template. Please advise if I've done something incorrectly. Just a FYI, `filteredColumns` could be an empty array.
```hbs
      {{#sortable-group tagName='ul' model=filteredColumns class='column-list value editable' onChange=(action 'reorderColumns') as |group|}}
        {{#each group.model as |column|}}
          {{#group.item class='editable' model=column as |columnItem|}}
            {{#form-button class='remove-meta-key' defaultAction=(action 'removeFromColumns' column)}}
              {{x-icon name='subtract-circle-1'}}
            {{/form-button}}
            <span class='field'>{{column.field}}</span>
            <span title={{column.title}} class='title'>{{column.title}}</span>
            {{#columnItem.handle class='thumb'}}
              {{x-icon name='navigation-drawer-1'}}
            {{/columnItem.handle}}
          {{/group.item}}
        {{/each}}
      {{/sortable-group}}
```

I'm able to bypass the error if I modify sortable-item's _tellGroup function like so:
```js
  _tellGroup(method, ...args) {
    let group = this.get('group');

    if (group && group[method]) { // <-- add additional check
      group[method](...args);
    }
  },
```

gtb104:
I forgot to mention, the `method` it can't find is `registerItem`.
![image](https://user-images.githubusercontent.com/1440968/68128084-0ec0e800-fee5-11e9-9a31-d87addc53350.png)

ygongdev:
Sounds like this is failing for you on the initial load? That seems very odd as registerItem is definitely a method on the sortable-group

BnitoBzh:
Fails for me too, same issue. Ember 3.13.x
_This v2.x seems to be be very unstable_

BnitoBzh:
Why not pass group's methods directly in the yield definition ?
With this behavior all `group=group` can be removed (from `sortable-item` calls)

BnitoBzh:
[...]

PR #309, opened 2019-11-04
Remove group=group from README

ygongdev:
Addresses the confusion regarding `group=group` in #306


AlexMayants:
Seems like the [migration guide](/adopted-ember-addons/ember-sortable/blob/master/MIGRATION_GUIDE_V2.md) needs a similar fix too.

Issue #310, opened 2019-11-05
Computed .volatile() deprecations


lukebenimble:
Just trying to tidy up my deprecation warnings and currently there are multiple computed property .volatile() warnings emanating from components/sortable-item.

Not a breaking issue at the moment, but will become one eventually - https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile

Happy to help with this, let me know if you'd like me to submit a PR.

Thanks for the great addon.

ygongdev:
Hey an effort was made in `1.x.x` in #276 
Feel free to tackle this via a PR for `2.x.x`

ygongdev:
Should be handled in `2.0.3` release.

PR #322, opened 2019-11-08
Deprecating volatile in favor of native getter and setter

ygongdev:
Giving it a shot.
Had to use `defineProperty` workaround to ensure compatibility with consumers that  do not have ember versions `3.10` and up.

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -1,6 +1,6 @@
 import { A } from '@ember/array';
 import Component from '@ember/component';
-import { set, get, computed } from '@ember/object';
+import { set, get, computed, defineProperty } from '@ember/object';
 import { run } from '@ember/runloop';
 import layout from '../templates/components/sortable-group';
 import {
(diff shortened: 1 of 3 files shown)


addon/components/sortable-group.js
```diff
@@ -674,5 +651,38 @@ export default Component.extend({
    */
   activateKeyDown(){
     this.set('isKeyDownEnabled', true);
-  }
+  },
+
+  /**
+   * Defining getters and setters to support native getter and setters until we decide to drop support for ember versions below 3.10
+   */
+  _setGetterSetters() {
```
lifeart: will it work defined in object itself like `get itemPosition() { ... }`?

ygongdev: I think so, the ember try passed when it used to fail.
@pzuraq?

ygongdev: @lifeart oh wait sorry I misread, directly using getter and setter like that is backward incompatibility. I originally tried that

knownasilya: You could maybe convert the component to an es class and then use getters. Not sure what the support matrix is though

ygongdev: @knownasilya At that point, I think I would just be doing the Octane migration then 😆 
I think we need to decide which point in the future, we should just outright deprecate all incompatible Ember versions.

knownasilya: According to the blog post you can get 3.4+ with a polyfill. Glimmer will be 3.14 which will leave many behind.
https://blog.emberjs.com/2019/01/26/emberjs-native-class-update-2019-edition.html

ygongdev: I was looking at the polyfill, but to my understanding I don't think it'll solve this particular issue right? Just by using the polyfill doesn't mean that I can now use native getter and setter in 3.4+? It's more of an effort to ease the migration of the entire class to native class, so it would probably not do anything regarding deprecating `volatile` and etc?

ygongdev: Or maybe you were just talking with regards to the future plans to move to native classes and not this particular problem 😄

ygongdev:
Related PR: #276 
Issues: #310

PR #326, opened 2019-11-18
Add beforedragend callback to `drag` helper

bantic:
👋 Thanks for maintaining this addon!

We recently ran into a case where we wanted to test application state while a drag was in-progress, before the final 'mouseup' was triggered. Would you accept this PR that adds a `beforedragend` optional callback?


ygongdev:
Hmm, looks fine to me.

Issue #231, opened 2019-02-15
drag/drop not resolving promise on drag end


apellerano-pw:
**Describe the bug**
Dragging a `sortable-item` and then releasing it does not cause `_complete()` to be called.

**To Reproduce**
1. Create some sortable items with an `onDragStop` action
2. Drag an item to a new spot and release mouse button
3. `onDragStop` action is never called

**Expected behavior**
`onDragStop` action is called

**Additional context**
I stepped thru the code and this seems to be the relevant bit:

```js
this.element.addEventListener('transitionend', deferred.resolve);
  transitionPromise = deferred.promise.finally(() => {
  this.element.removeEventListener('transitionend', deferred.resolve);
});
```
_(from https://github.com/heroku/ember-sortable/blob/master/addon/mixins/sortable-item.js#L665-L668)_

That `transitionend` event never fires, so the promise never resolves.

As a workaround I was able to manually set `isAnimated=false` on the sortable-item and then it works again, since it uses a `run.later` to resolve the promise instead of the `transitionend` event.

apellerano-pw:
I believe this is relevant to the conversation in this PR https://github.com/heroku/ember-sortable/pull/223#issuecomment-460650716

fivetanley:
Maybe `isAnimated` should have a default of `false`? We don't include any styles for animations by default it seems.

fivetanley:
It looks like we do check to make sure the element is actually animated. Do you have any custom styles for transitions that you're using? https://github.com/heroku/ember-sortable/blob/master/addon/mixins/sortable-item.js#L129

apellerano-pw:
Doesn't look like we are specifying any transition properties on the element. Here's a snapshot from `getComputedStyle(el)`, lmk if seeing more of it would be useful:
```js
transition: "all 0s ease 0s"
transitionDelay: "0s"
transitionDuration: "0s"
transitionProperty: "all"
transitionTimingFunction: "ease"
```

`isAnimated` will check for `transition-property: all` and that appears to be the default value (https://developer.mozilla.org/en-US/docs/Web/CSS/transition-property). Perhaps `transition-duration` and/or `transition-delay` would be a better indicator of whether to listen for `transitionend`?

apellerano-pw:
If I change the implementation of `isAnimated` to do
```js
let time = parseInt(getComputedStyle(el).transitionDuration, 10);
return time > 0;
```

[...]

PR #343, opened 2019-12-16
Added check for transitionDuration to isAnimated

cah-briangantzler:
Fixes https://github.com/adopted-ember-addons/ember-sortable/issues/231

### DIFF
--- a/addon/components/sortable-item.js
+++ b/addon/components/sortable-item.js
@@ -627,7 +627,7 @@ export default Component.extend({
         let el = this.element;
         let property = getComputedStyle(el).transitionProperty;
 
-        return /all|transform/.test(property);
+        return /all|transform/.test(property) && this.get("transitionDuration") > 0;
       }
     });
 
--- a/tests/acceptance/smoke-test.js
+++ b/tests/acceptance/smoke-test.js
@@ -91,7 +91,28 @@ module('Acceptance | smoke', function(hooks) {
     assert.equal(scrollableContents(), 'Tres Dos Uno Cuatro Cinco');
   });
 
-  test('reordering with touch events', async function(assert) {
+  test('Test is Animated still works without css for transitionDuration', async function(assert) {
+    await visit('/');
+
+    assert.equal(verticalContents(), 'Uno Dos Tres Cuatro Cinco');
+    assert.equal(horizontalContents(), 'Uno Dos Tres Cuatro Cinco');
+    assert.equal(tableContents(), 'Uno Dos Tres Cuatro Cinco');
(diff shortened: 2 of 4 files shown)


ygongdev:
@cah-briangantzler, looks good so far! Just a couple of questions from my end.

1. Are we able to visually confirm that this fixes #231? e.g it works after removing the transition from css.
2. I see that you write a nice unit test for `isAnimated`, which is awesome. Are we able to take it one step further and write an integration or smoke test that fails #231 without your change and passes with your change?

Just to ensure that we have this case covered in the future.
Thanks for your hard work!

cah-briangantzler:
I actually did have that as an acceptance  test and then removed it. The problem was from the test it was not obvious what and how the test was working. Second for a failure, the promise never returns, so the test just stalled, the promise never resolved. Wasnt sure how to timeout waiting on a promise and fail the promise.

So long or short, I had a test, it stalled, when I put the fix in it completed normally.

cah-briangantzler:
Also, now that Im thinking about it, you cant run a test with the change and a test without the change,. The code is either there or it isnt. This fix is to stop a bad thing from happening. Once its fixed, you cant make the bad thing happen, thats the point, so dont know you can make a failing test

ygongdev:
Oh, sorry I should make it clear. I didn't expect to have both failing test and working test.
I was hoping that the current passing test after your change was failing before your change, so it verifies that the bug has been fixed.

cah-briangantzler:
Yes. It wasnt so much failing as no completing, but yes I did have that. But it wasnt really obvious what was being tested. Can add it back easy enough. An then you can decide if it actually conveys anything

cah-briangantzler:
Just added and pushed. If you clone repo, go into this branch, undo the one line in sortable-item and run the new test, it will hang, put the line back and it will complete. But the test itself looks like a duplicate and its not immediately obvious what the test is doing. It just looks like an extra test. Added a few comments to help.

cah-briangantzler:
Ill take a look at this later tonight (EST here). These tests pass locally. The test I wrote is passing now, but another test is failing that shouldnt be effected. Will check it out

ygongdev:
LGTM, thanks!

sean-realestate:
Can you guys make a new publish base on this commit please? Seems like the latest version does not have this change.

ygongdev:
@sean-realestate sure, but just want to confirm if this PR will address your use case?

sean-realestate:
> @sean-realestate sure, but just want to confirm if this PR will address your use case?

I believe so. Version 2.1.1 does not work for me. After I use the change of this PR, the addon worked.

ygongdev:
@sean-realestate 👍 I just released `2.1.2` with this change.

PR #345, opened 2019-12-19
Create a modifier version - Part 2 - more streamlined

cah-briangantzler:
Provides for a more streamlined version of the modifier

```
<ol {{sortable-group onChange=this.update a11yAnnouncementConfig=this.myA11yConfig}}>
  {{#each model.items as |item|}}
    <li {{sortable-item model=item}}>
      {{item.name}}
      <span class="handle" {{sortable-handle}}>&varr;</span>
    </li>
  {{/each}}
</ol>
```

This implements the purposed issue https://github.com/adopted-ember-addons/ember-sortable/issues/342


st-h:
@cah-briangantzler Is there anything I can help you with here? Just wanted to add this Addon to my app, but noticed it's either refactoring quite a bit of my markup, or making modifiers available.

cah-briangantzler:
As far as Im concerned this is completed. Just waiting on them to approve. The modifier version is only back to 3.8, the component version is back farther. I updated the travis to no longer test the older versions.

The tests are failing on the beta and canary (Both component and modifier), I dont know if thats a blocker to releasing this.

If you were refactoring anyway, and on 3.8+ the modifier is definitely the way to go. Gives you more flexibility. We use ember-yeti-table and wanted to make the rows use this addon. We cant with the component version, but can apply the modifiers to the yeti-table component that represents the tr.

addon/modifiers/sortable-item.js
```diff
@@ -0,0 +1,762 @@
+import Ember from 'ember';
```
ygongdev: Just one minor thing, can we add the change from #347 into the modifier as well?
It'll keep both component and modifier in sync when I merge this.

cah-briangantzler:
Yea, Ill get to that tonight.

Future work would probably be to extract some of this methods into a utility object that could be shared between the two to help with future maintenance.

addon/modifiers/sortable-item.js
cah-briangantzler: Yea Ill get to that tonight.

Should in future extract some of those methods to a util that could be shared for better maintenance

ygongdev: Agreed, once this change is in, any latter improvements should be easily streamline-able. 
After this, my plan is to release a `beta` for anyone to try out the `modifier` and to give us ample time to get and fix bug reports. Eventually, we should keep the `modifier` and kill off the `component` in a `v3` release.

cah-briangantzler:
Im not sure whats causing the 'no environment variable set' job to be created thats failing. All the tests pass except beta and canary. The travis is set to allow canary to fail, but not beta

ygongdev:
The tests beta and canary tests are also failing on master, so I'm thinking something is going on with `3.17+`, but I'm ok as long as `3.15` is fine

chriskrycho:
👏 thanks for driving this, @cah-briangantzler!

PR #349, opened 2020-02-04
fixes handle when using modifiers

st-h:
I noticed that the handle does not work correctly within the modifiers version: Even when it is present, the whole sortable element is draggable. We should probably still add some tests that try to sort by trying to move a sortable element where a handler is present.

### DIFF
--- a/addon/modifiers/sortable-item.js
+++ b/addon/modifiers/sortable-item.js
@@ -266,10 +266,7 @@ export default class SortableItemModifier extends Modifier {
       return;
     }
 
-    let handle = this.handle;
-
-    // todo what does this do? why do you have to null handle
-    if (!this.handleElement && !startEvent.target.closest(handle)) {
+    if (this.handleElement && !startEvent.target.closest(this.handle)) {
       return;
     }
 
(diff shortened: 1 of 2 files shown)


ygongdev:
Thanks for the PR and yes I agree that we should add some tests.

Is it possible to add a test to assert that sorting using the entire element fails when there is a handle?
I think it should be do-able via the `reorder` test helper.

st-h:
Yes, I think there shouldn't be an issue to add those tests. I just don't have the time right now... sorry.

st-h:
@ygongdev I have added a test to assert previously broken behaviour. Are there any other issues keeping you from merging this?

ygongdev:
Sorry busy week. Left a comment

addon/modifiers/sortable-item.js
```diff
@@ -743,12 +740,11 @@ export default class SortableItemModifier extends Modifier {
     // Instead of using `event.preventDefault()` in the 'primeDrag' event,
```
ygongdev: So the idea is that if a handle isn't specified, the entire element itself is draggable by default. To disable dragging in general would be done via `isDraggingDisabled`. Just want to make sure that behavior is retained.

st-h: This PR contains no such changes. It only fixes the handle in the modifier version, which has been broken compared to the non-modifier version. This is a mere bug fix and should not introduce any change in behaviour. The logic is identical to the non modifier version now. Looks like this was work in progress previously, as the comment I removed indicated: 

> // todo what does this do? why do you have to null handle

mike183:
Any chance of getting this merged? I have been using it in our application for a couple of weeks without any issues.

ygongdev:
@mike183 apologies forgot about this PR 😬

mike183:
No worries, thanks for merging 😁 🎉

PR #352, opened 2020-02-07
wip typescript support

st-h:
Some things should sort itself out once https://github.com/ember-modifier/ember-modifier/pull/9 is merged.

It would be terrific if someone with more typescript experience could have a look at this.


chriskrycho:
@st-h you should be able to land this on top of a new major versions of ember-modifier which includes types. I’m happy to help work through how to land the TS support here once we finish getting that across the line!

st-h:
@chriskrycho thanks for your reply. There probably could have been much better timing to try to get a migration to ts out the door. Will take a stab at it later again. Probably will still have some questions then :)

chriskrycho:
Totally understand! I was buried in Ember upgrades for the app I work on back in February, and haven’t really had bandwidth till the last two weeks. Let me know if/when you do want to take another swing at it!

st-h:
@chriskrycho no problem. I was referring to the current state of this addon, where the move to modifiers is still in progress. Should wait for this to land, otherwise adding TS could easily mean maintaining another branch for some time. And TS support should be better by then, so it pretty much just was bad timing when I tried to convert the addon. Appreciate your help - as will all TS questions 😊

PR #353, opened 2020-02-07
modifier: handle renders as draggable when dragging is disabled

st-h:
For the modifier version there currently wasn't a way for a draggable elements to know when dragging is disabled. This PR adds the attribute `data-sortable-disabled=true` to the sortable item.
Additionally event listeners are not added to disabled elements. For instance we have a list, that usually is fixed, but want to allow for editing by resource owners. Always attaching event listeners would be a waste of resources in this case.
Event listeners are now attached to the handle instead of the sortable item - just seemed to make sense, as the item itself would not be draggable, only the handle would be

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ $ ember install ember-sortable
 
 ## Usage
 
-### component 
+### component
 ```hbs
 {{! app/templates/my-route.hbs }}
 
@@ -47,7 +47,7 @@ $ ember install ember-sortable
 {{/sortable-group}}
 ```
 
-### modifier 
+### modifier
 ```hbs
 {{! app/templates/my-route.hbs }}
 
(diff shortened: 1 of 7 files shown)


cah-briangantzler:
Hmmm, looks like there were no actual tests in the component version checking for the isDraggingDisabled. I thought once all the tests passed I had fixed everything. Good catch.

kiwiupover:
@st-h thanks steve for your work on this PR. Are you still interested in pushing it over the line?

st-h:
@kiwiupover honestly, I totally forgot that I even created this PR. However, I still need the feature to completely disable the sortable list. Currently I maintain a regular list, which is replaced with a sortable list when needed in order to avoid having useless event listeners. I'll try to follow up with this when I have time. Is there anything missing, or is it just the conflicts that now prevent merging?

st-h:
@kiwiupover After looking at the current code, some of the changes I made back in the days are no longer necessary. I opened a new PR to address the remains issues: #485 
Could you please have a look if anything is missing to get this merged?

PR #354, opened 2020-02-08
sortable-group breaks when new elements are added to the list after it has been initialized

### LINKED ISSUE #351: [2.2.0-beta.0] Modifiers: Breaks when items are loaded asynchronously
When using the new modifiers syntax, it seems to break with several errors being thrown if the array of items being sorted was loaded asynchronously.

I came across this bug when implementing a component that fetches its data outside of the normal route lifecycle.

After a small amount of debugging, I found that due to the data being fetched asynchronously, [this line](https://github.com/adopted-ember-addons/ember-sortable/blob/da8107043f78826e35871228a99c0cd7e8af0717/addon/modifiers/sortable-group.js#L714) was returning an empty NodeList.

Since the `sortable-group` modifier is unaware of the items being used to render the `sortable-item` elements, it isn't able to recognise that it needs to re-initialize itself.

Not really sure on the best approach for resolving this, any help would be appreciated 🙂

**To Reproduce**
1. Create a component that fetches data asynchronously
2. Render the items using code similar to the below
3. Attempt to reorder list of items
4. Observe errors in console (errors attached at bottom of issue)

```hbs
<ol {{sortable-group onChange=this.reorderItems}}>
  {{#each this.asyncFetchedItems as |item|}}
    <li {{sortable-item model=item}}>{{item.name}}</li>
  {{/each}}
</ol>
```

**Expected behavior**
`{{sortable-group}}` modifier should be able to handle `{{sortable-item}}` elements being rendered asynchronously.

**Additional context**
```
Ember: v3.15.0
EmberCLI: v3.15.1
Node: v12.14.0
```

I've attached the errors that were thrown below,
```
sortable-item.js:370 Uncaught TypeError: Cannot read property 'prepare' of undefined
    at SortableItemModifier._startDrag (sortable-item.js:370)
    at SortableItemModifier._prepareDrag (sortable-item.js:335)
    at Backburner._run (backburner.js:1013)
    at Backburner._join (backburner.js:989)
    at Backburner.join (backburner.js:760)
    at join (index.js:179)
    at index.js:274
```
```
backburner.js:389 Uncaught Error: You attempted to schedule an action in a queue (actions) for a method that doesn't exist
    at DeferredActionQueues.schedule (backburner.js:389)
    at Backburner._scheduleExpiredTimers (backburner.js:1122)
    at Backburner._runExpiredTimers (backburner.js:1095)
```
```
sortable-item.js:595 Uncaught TypeError: Cannot read property 'update' of undefined
    at SortableItemModifier._drop (sortable-item.js:595)
    at sortable-item.js:362
    at Backburner._run (backburner.js:1013)
    at Backburner.run (backburner.js:754)
    at run (index.js:125)
    at drop (sortable-item.js:361)
```
```
backburner.js:389 Uncaught Error: You attempted to schedule an action in a queue (actions) for a method that doesn't exist
    at DeferredActionQueues.schedule (backburner.js:389)
    at Backburner._scheduleExpiredTimers (backburner.js:1122)
    at Backburner._runExpiredTimers (backburner.js:1095)
```

st-h:
fixes #351

### DIFF
--- a/addon/modifiers/sortable-group.js
+++ b/addon/modifiers/sortable-group.js
@@ -711,16 +711,17 @@ export default class SortableGroupModifier extends Modifier {
   }
 
   didReceiveArguments() {
+    this.initializeGroup();
+  }
+
+  initializeGroup() {
     let items = this.element.querySelectorAll("[data-sortable-item]");
 
     // create the array of modifiers that match the items. If items are changed, a property needs updated
     this.items = Array.from(items).map(item => {
-      item.sortableItem.sortableGroup = this;
+      set(item.sortableItem, 'sortableGroup', this);
       return item.sortableItem;
     });
-  }
-
-  didUpdateArguments() {
   }
 
   didInstall() {
@@ -728,6 +729,15 @@ export default class SortableGroupModifier extends Modifier {
 
     this.announcer = this._createAnnouncer();
     this.element.insertAdjacentElement('afterend', this.announcer);
+
+    var observer = new MutationObserver(() => {
+      run(() => this.initializeGroup());
+    });
+    observer.observe(this.element, {
+      childList: true,
+      attributes: false,
+      subtree: false
+    });
   }
 
   willRemove() {
--- a/addon/modifiers/sortable-group.js
+++ b/addon/modifiers/sortable-group.js
@@ -730,10 +730,10 @@ export default class SortableGroupModifier extends Modifier {
     this.announcer = this._createAnnouncer();
     this.element.insertAdjacentElement('afterend', this.announcer);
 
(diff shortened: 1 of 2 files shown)


cah-briangantzler:
If I understand the MutationObserver correctly its going to fire is anything is changed in the DOM anywhere.

The sortable group looks down in its DOM to find the items. I would suggest just changing the items to look up in the DOM to find the sortable group and register themselves. I didnt do this because it was more difficult and didnt think it was needed. Given your problem I can see how it might be.. What do you think about this approach?

st-h:
closing in favour of scheduling on the run loop instead of observing the dom

PR #355, opened 2020-02-09
Have sortable-item register/deregister with the group instead of the …

bgantzler:
…group looking up items on install

Attempts to fix https://github.com/adopted-ember-addons/ember-sortable/issues/351

### DIFF
--- a/addon/modifiers/sortable-group.js
+++ b/addon/modifiers/sortable-group.js
@@ -564,7 +564,7 @@ export default class SortableGroupModifier extends Modifier {
     const index = this.items.indexOf(item);
     if (index !== -1) {
       const items = [...this.items.slice(0, index), ...this.items.slice(index + 1)];
-      this.set("items", items);
+      set(this, "items", items);
     }
   }
 
@@ -711,13 +711,6 @@ export default class SortableGroupModifier extends Modifier {
   }
 
   didReceiveArguments() {
-    let items = this.element.querySelectorAll("[data-sortable-item]");
-
-    // create the array of modifiers that match the items. If items are changed, a property needs updated
-    this.items = Array.from(items).map(item => {
-      item.sortableItem.sortableGroup = this;
-      return item.sortableItem;
-    });
   }
 
   didUpdateArguments() {
@@ -728,6 +721,7 @@ export default class SortableGroupModifier extends Modifier {
 
     this.announcer = this._createAnnouncer();
     this.element.insertAdjacentElement('afterend', this.announcer);
+    this.element.sortableGroup = this;
   }
 
   willRemove() {
--- a/addon/modifiers/sortable-item.js
+++ b/addon/modifiers/sortable-item.js
@@ -751,10 +751,26 @@ export default class SortableItemModifier extends Modifier {
     this.addEventListener();
     this.element.dataset.sortableItem=true;
     this.element.sortableItem = this;
+
+    // Find my group
+    // we have to schedule because the items are installed before the group
+    run.schedule('afterRender', () => {
+      let parent = this.element.parentElement;
+      while (parent) {
+        if (parent.sortableGroup) {
+          parent.sortableGroup.registerItem(this);
+          this.sortableGroup = parent.sortableGroup;
+          break;
+        }
+        parent = parent.parentElement;
+      }
+    });
   }
 
   willRemove() {
     this.removeEventListener();
+    this.sortableGroup.deregisterItem(this);
+    delete this.sortableGroup;
   }
 
 }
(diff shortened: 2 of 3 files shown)


cah-briangantzler:
Is there anyone that can review and approve this?

addon/modifiers/sortable-item.js
```diff
@@ -755,10 +755,26 @@ export default class SortableItemModifier extends Modifier {
...
+      let parent = this.element.parentElement;
+      while (parent) {
+        if (parent.sortableGroup) {
+          parent.sortableGroup.registerItem(this);
+          this.sortableGroup = parent.sortableGroup;
+          break;
+        }
+        parent = parent.parentElement;
+      }
+    });
```
ygongdev: Ah, this makes me kind of sad.
Ember-sortable was originally implemented as separate entities that did not know  about each other without explicit metadata to track, e.g `sortable-group` and `sortable-item`. The ability to create higher order components via `yield` was on the major improvements done in `2.x` because it allowed us to get rid of most of the metadata needed to always keep the two connected. In almost all cases, you are using `sortable-group` and `sortable-item` together, so the tight coupling was warranted.

I wonder if there's a way we can do the same for `modifiers` in a higher order fashion.

cah-briangantzler: There was talk of a `modifier` helper similar to the `component` helper which would have definitely solved the concern here. Sadly that helper never came to be.

The main problem with the components was it as a fixed DOM element. Only changeable by tagName (going away with glimmer) and not generically useable by other addons components. Any element should be sortable, even elements of other components. Say there is another addon that does a really great job of ol/li management. We can not user ember-sortable with that add on because the sortable components insist on managing the DOM. . But with modifiers,  we can apply the modifiers to the other addons components and now those components are also sortable (This does require the other addon to use `...arguments` in the correct place. An easy fix and almost a guarantee when converted to glimmer).

Without a `modifier` helper, the only way to make modifiers exhibit the shared state is to require the consumer to assign the state on each modifier, a burden I thought needless when it could be derived for the consumer.

cah-briangantzler: Oh, FYI the above code exists in the group of the already approve code that is in the beta. This is just moving the code from the group to the item to support the async load.

addon/modifiers/sortable-group.js
```diff
@@ -711,13 +711,13 @@ export default class SortableGroupModifier extends Modifier {
...
-
-    // create the array of modifiers that match the items. If items are changed, a property needs updated
-    this.items = Array.from(items).map(item => {
-      item.sortableItem.sortableGroup = this;
-      return item.sortableItem;
-    });
+    // let items = this.element.querySelectorAll("[data-sortable-item]");
+    //
+    // // create the array of modifiers that match the items. If items are changed, a property needs updated
+    // this.items = Array.from(items).map(item => {
```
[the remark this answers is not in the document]
cah-briangantzler: Yes. This was the code that that was in the group and now is effectively moved to the item. Basically the below code is replacing this.

I have created another commit and pushed to remove this

addon/modifiers/sortable-item.js
ygongdev: I agree that modifier does bring a whole new perspective into the ecosystem. We will need to re-visit the architecture and design.
But as to unblock current tester for modifiers, are there tests that we can write to ensure that this `async` behavior is resolved?

cah-briangantzler: https://github.com/st-h created a test in his PR https://github.com/adopted-ember-addons/ember-sortable/pull/354 which was closed in favor of this one. I copied that test over and its in this PR. It added items to the array after render and checks that the elements are added on screen and are then sortable

ygongdev: sounds good.

tests/integration/modifiers/sortable-group-test.js
```diff
@@ -0,0 +1,74 @@
...
+
+module('Integration | Modifier | sortable-group', function(hooks) {
+  setupRenderingTest(hooks);
+
+  // Replace this with your real tests.
+  test('it renders', async function(assert) {
+    await render(hbs`<div {{sortable-group}}></div>`);
+
+    assert.ok(true);
+  });
```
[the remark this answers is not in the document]
cah-briangantzler: Thats weird. I do not see this test in my branch.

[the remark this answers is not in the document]
cah-briangantzler: Was looking in wrong place. This file/tests was what I copied over from st-h. Yea looking at it, should be removed

cah-briangantzler:
Where are we sitting with this now? I know there was some concern with the after render schedule but this does work and I though we were going to get this merged in to allow further testing.

In the mean time I have learned a couple things that will alleviate some of my concerns with using a service. If there is still a desire to to have a service be the state manager and NOT have the items register themselves with the group I have an idea on how to do that now (without burdening the user api interaction). Is this something that is desired?

bgantzler:
There is another PR https://github.com/adopted-ember-addons/ember-sortable/pull/361 that maintains state via a service instead of within the group and using after render to register. These are competeing PRs. Choose the one you want and close the other

ygongdev:
Closing this in favor of the service implementation

PR #357, opened 2020-02-18
Actually use `onDragStop` argument rather then re-using `onDragStart`.

alexabreu:
Noticed that the supplied `onDragStart` handler was getting called twice and this seems to explain why. Looks like a simple refactoring mistake which slipped through in the previous beta release.


miguelcobain:
Unfortunately there is no release on npm with this fix. :(
Would it be possible to release soon?

PR #360, opened 2020-02-28
Fix/auto scroll with document container

### LINKED ISSUE #233: Scroll-on-edges is not working when the scroll container is document
**Describe the bug**
Scroll-on-edges is not working.  When I have a list that overflows the height of the page and the scroll container is the `document` (no internal scrolling), dragging a list item to the top or bottom is not causing the list to scroll.  I have created an ember-twiddle to demonstrate the problem...

**To Reproduce**
1. Go to https://ember-twiddle.com/431624b67579f1a7c25b0675b682c58b?openFiles=templates.application.hbs%2C
2. try to drag a list item to the bottom (or top) edges of the list

**Expected behavior**
The list should scroll to allow dragging the list item to the very bottom, or top of the list

**Additional context**
ember-sortable v1.12.4
This works with 1.12.1

romgere:
This PR aims to fix #233

The "real" fix concerns only the file [system/scroll-container.js](https://github.com/adopted-ember-addons/ember-sortable/blob/762761e37030fcdaa76aa50c65523d3fa062b9c9/addon/system/scroll-container.js)

I wrote a new route in dummy app to allow some acceptance tests. I also added a new argument on `sortable-item` component & modifier to enable the check about scroll bounds when runing test suite.

& last, the acceptance test about "auto scroll" when container is the document includes a hack to "unlock" the height of `#ember-testing` div and allowing scrolling on the document element. 
I personnaly dislike this kind of "hack", but I didn't found any other way to test it.
If someone have an idea... I'm all ears.

### DIFF
--- a/addon/components/sortable-item.js
+++ b/addon/components/sortable-item.js
@@ -148,6 +148,9 @@ export default Component.extend({
   */
   _direction: computed.readOnly('direction'),
 
+
+  disableCheckScrollBounds: Ember.testing,
+
   init() {
     this._super(...arguments);
     this._setGetterSetters();
@@ -425,8 +428,7 @@ export default Component.extend({
         requestAnimationFrame(checkScrollBounds);
       }
     };
-
-    if (!Ember.testing) {
+    if (!this.disableCheckScrollBounds) {
       requestAnimationFrame(checkScrollBounds);
     }
   },
--- a/addon/components/sortable-item.js
+++ b/addon/components/sortable-item.js
@@ -1,4 +1,3 @@
-import Ember from 'ember';
 import { Promise, defer } from 'rsvp';
 import Component from '@ember/component';
 import { computed, defineProperty } from '@ember/object';
@@ -11,6 +10,11 @@ import { getBorderSpacing } from '../utils/css-calculation';
 import { DRAG_ACTIONS, ELEMENT_CLICK_ACTION, END_ACTIONS } from '../utils/constant';
 import { getX, getY } from '../utils/coordinate';
 import { buildWaiter } from 'ember-test-waiters';
+import config from 'ember-get-config';
+const { environment } = config;
+const isTesting = environment === 'test';
+
+
 
 const sortableItemWaiter = buildWaiter("sortable-item-waiter");
 
@@ -149,7 +153,7 @@ export default Component.extend({
   _direction: computed.readOnly('direction'),
 
 
-  disableCheckScrollBounds: Ember.testing,
+  disableCheckScrollBounds: isTesting,
 
   init() {
     this._super(...arguments);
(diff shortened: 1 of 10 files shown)


romgere:
Anybody to give me a feedback for this ? I really liked to see this released in a future release

addon/components/sortable-item.js
```diff
@@ -148,6 +148,9 @@ export default Component.extend({
   */
   _direction: computed.readOnly('direction'),
 
+
+  disableCheckScrollBounds: Ember.testing,
```
Gaurav0, line 152: Ember.testing should not be used. See discussion here: https://github.com/emberjs/rfcs/pull/619#issuecomment-618789784

addon/modifiers/sortable-item.js
```diff
@@ -191,6 +191,16 @@ export default class SortableItemModifier extends Modifier {
...
   isBusy;
 
+  /**
+   Removes the ability for the current item to be dragged
+   @property isDraggingDisabled
+   @type  boolean
+   @default false
+   */
+  get disableCheckScrollBounds() {
+    return this.args.named.disableCheckScrollBounds != undefined ? this.args.named.disableCheckScrollBounds : Ember.testing;
```
Gaurav0, line 201: same issue.

addon/components/sortable-item.js
romgere: I can not use the solution you suggest with `import config from 'my-application/config/environment';` in an add-on :thinking: What do you suggest here ? Using `ember-get-config`?

Gaurav0: I think that would be fine.

romgere: OK, it's done.

addon/modifiers/sortable-item.js
romgere: Done here also...

btecu:
What's the status here? It would be nice to get the scroll fixed.

ygongdev:
retriggering ci

romgere:
I rebased on master to try to fix the CI..

CI is OK on local (with ember-try on ember 3.18) :thinking: IDK why the CI is failing here, anyone to help ?

Gaurav0:
ember-try configuration should be for 3.16 and 3.20, not 3.15 and 3.18. Let me see if that fixes it...

Gaurav0:
Nope. It looks like master branch is failing tests.

ygongdev:
@Gaurav0 any tips on ember try? Is there a standard to check which version to test? For example, why 3.16 and 3.20 instead of 3.15 and 3.18? 

After playing around with different versions, it always looks like the CI fails on version 3.17 and 3.18. So maybe it is a bug in those versions? I'm not too concerned about CI failing for specific versions b/c that seems like something is wrong with those ember source versions, especially since ember-release, ember-beta, and ember-canary passes

ygongdev:
@romgere merged, I'm ignoring the CI for now since it only happens on 1 or 2 intermediate ember versions and we can fix forward later since it doesn't seem related to your PR.

ygongdev:
@Gaurav0 nvm, I think I got it. Is it because those versions are LTS?

PR #361, opened 2020-03-01
Use a service to track group/item state

bgantzler:
This PR uses a service instead of item after render to register and maintain state.

This does require that is there is more than one group in the DOM that both the group and items contain an additional param of the group name.

### DIFF
--- a/tests/dummy/app/templates/modifier.hbs
+++ b/tests/dummy/app/templates/modifier.hbs
@@ -33,10 +33,11 @@
       <ol data-test-vertical-demo-group
         {{sortable-group
           onChange=this.updateDifferentSizedModels
+          groupName="verticle-size"
         }}
       >
         {{#each this.differentSizedModels as |item|}}
-          <li data-test-vertical-demo-item class="word-break" {{sortable-item model=item}}>
+          <li data-test-vertical-demo-item class="word-break" {{sortable-item model=item groupName="verticle-size"}}>
             <label for="demo-input">{{item}}</label>
             <input id="demo-input" type="text">
             <span data-item={{item}} data-test-vertical-demo-handle class="handle" {{sortable-handle}}>
@@ -56,10 +57,11 @@
           onChange=this.update
           itemVisualClass=this.itemVisualClass
           handleVisualClass=this.handleVisualClass
+          groupName="horizontal"
         }}
       >
         {{#each this.model.items as |item|}}
-          <li data-test-horizontal-demo-handle tabindex={{"0"}} {{sortable-item model=item}}>
+          <li data-test-horizontal-demo-handle tabindex={{"0"}} {{sortable-item model=item groupName="horizontal"}}>
             {{item-presenter item=item}}
           </li>
         {{/each}}
@@ -77,9 +79,9 @@
             <th>Item</th>
           </tr>
         </thead>
-        <tbody {{sortable-group onChange=this.update}}>
+        <tbody {{sortable-group onChange=this.update groupName="table"}}>
           {{#each this.model.items as |item|}}
-            <tr data-test-table-demo-item class="handle" {{sortable-item model=item }}>
+            <tr data-test-table-demo-item class="handle" {{sortable-item model=item groupName="table"}}>
               <td><span data-test-table-demo-handle data-item={{item}} class="handle">&vArr;</span></td>
               <td>{{item}}</td>
             </tr>
@@ -91,9 +93,9 @@
     <section class="vertical-spacing-demo">
       <h3>Vertical with 15px spacing</h3>
 
-      <ol {{sortable-group onChange=this.update}}>
+      <ol {{sortable-group onChange=this.update groupName="verticle-15"}}>
         {{#each this.model.items as |item|}}
-          <li {{sortable-item model=item spacing=15}} tabindex="0">
+          <li {{sortable-item model=item spacing=15 groupName="verticle-15"}} tabindex="0">
             {{item}}
             <span data-test-vertical-spacing-demo-handle data-item={{item}}>
             </span>
@@ -105,9 +107,9 @@
     <section class="vertical-distance-demo">
       <h3>Vertical with distance set to 15</h3>
 
-      <ol {{sortable-group onChange=this.update}}>
+      <ol {{sortable-group onChange=this.update groupName="verticle-15d"}}>
         {{#each this.model.items as |item|}}
-          <li data-test-vertical-distance-demo-handle {{sortable-item model=item distance=15}} tabindex="0">
+          <li data-test-vertical-distance-demo-handle {{sortable-item model=item distance=15 groupName="verticle-15d"}} tabindex="0">
             {{item}}
             <span data-item={{item}}>
             </span>
@@ -120,9 +122,9 @@
       <h3>Scrollable</h3>
 
       <div class="sortable-container">
-        <ol {{sortable-group onChange=this.update}}>
+        <ol {{sortable-group onChange=this.update groupName="scrollable"}}>
           {{#each this.model.items as |item|}}
-            <li data-test-scrollable-demo-handle {{sortable-item model=item}}>
+            <li data-test-scrollable-demo-handle {{sortable-item model=item groupName="scrollable"}}>
               {{item}}
               <span class="handle" data-item={{item}} {{sortable-handle}}>
                 <span>&vArr;</span>
@@ -148,21 +150,18 @@
   <section class="vertical-demo">
     <h3>Vertical</h3>
 
-    {{#sortable-group
-            data-test-vertical-demo-group-no-css
-            handleVisualClass=handleVisualClass onChange=(action "update")
-            model=model.items as |group|
-    }}
-      {{#each group.model as |item|}}
-        {{#group.item data-test-vertical-demo-item-no-css tagName="li" model=item  as |groupItem|}}
+    <ol {{sortable-group data-test-vertical-demo-group-no-css
+                         onChange=this.update handleVisualClass=handleVisualClass
+                         groupName="verticle-no-css"}}>
+      {{#each this.model.items as |item|}}
+        <li data-test-scrollable-demo-handle-item-no-css {{sortable-item model=item groupName="verticle-no-css"}}>
           {{item}}
-          {{#groupItem.handle data-test-vertical-demo-handle-no-css class="handle"}}
-            <span data-item={{item}}>
-              <span>&vArr;</span>
-            </span>
-          {{/groupItem.handle}}
-        {{/group.item}}
+          <span class="handle" data-test-vertical-demo-handle-no-css data-item={{item}} {{sortable-handle}}>
+            <span>&vArr;</span>
+          </span>
+        </li>
       {{/each}}
-    {{/sortable-group}}
+    </ol>
+
   </section>
 </article>
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ $ ember install ember-sortable
       <span class="handle" {{sortable-handle}}>&varr;</span>
     </li>
   {{/each}}
-</ol>>
+</ol>
 ```
 
 The `onChange` action is called with two arguments:
@@ -123,7 +123,7 @@ The modifier version does not support `groupModel`, use the currying of `action`
       <span class="handle" {{sortable-handle}}>&varr;</span>
     </li>
   {{/each}}
-</ol>>
+</ol>
 ```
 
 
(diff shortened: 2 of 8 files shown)


roncodes:
I tried to run your branch on my local and received this error

```
Uncaught TypeError: this.registerItem is not a function
    at Class.didInsertElement (sortable-item.js:155)
```

cah-briangantzler:
Could you be more specific when you say run?

All the tests run in this branch. 

there are two ways to use this, as a component, as a modifier. This branch is providing the modifier support. It has not in any way changed the component version.

line 155 of sortable-item.js in either version does not contain this.registerItem. I do however find it at 162 of the component version in the didInsert method. If you are using the component version then the changes in this branch do nothing for you, and you should find the same error in the master branch. 

What version of ember are you using? as I also notice others are experiencing similiar problems https://github.com/adopted-ember-addons/ember-sortable/issues/363

mike183:
I've just tried this branch out and it seems to be working as expected in my application.

@bgantzler / @cah-briangantzler Any chance you'd be able to rebase it against master? There are some fixes missing which are currently in master, and I may have to run this branch while we wait for either it or #355 to be merged.

cah-briangantzler:
It appears to me the branch on my repository awaiting to be merged contains the latest commit on the master branch in the main repo.  What commits do you think you are missing?

mike183:
@cah-briangantzler You're right, I think I've actually just found another bug which is very similar to one that had recently been fixed.

Apologies!

ygongdev:
Sorry for the long delay @cah-briangantzler. I've not had much free time to dive into the PR as much as I would've liked.
Given that you have such passion for making `ember-sortable` great, would you like to be a contributor, so you have more power to drive this forward?

cah-briangantzler:
I dont have a problem with that, but going to need some hand holding at first as I am not very familiar with publishing things on npm and such.

For this particular PR, given that it does NOT modify the original components, I would most likley approve it and then fix any reported bugs. I believe there are several people already using the branch at this time.

ygongdev:
That's fine. @jgwhite are you onboard with adding @cah-briangantzler as a contributor? I don't have the permission to edit collaborators.

jgwhite:
@knownasilya could you help us out with the above request to add a maintainer?

ygongdev:
Bump! @knownasilya

btecu:
@MelSumner can also help adding the new contributor to the team.

knownasilya:
@cah-briangantzler you should be all set.

cah-briangantzler:
Im fine with the git portion, merging an such, but actually releasing master to npm, no so familiar with. There some docs or something about how to do that aspect?

knownasilya:
@cah-briangantzler what's your npm username?

cah-briangantzler:
I have never released an addon on npm so Im going to say I dont have one. You can send an email to brian.gantzler@cardinalhealth.com if you want to not clutter up this thread

tests/dummy/app/templates/modifier.hbs
```diff
@@ -148,21 +150,18 @@
...
+      {{#each this.model.items as |item|}}
+        <li data-test-scrollable-demo-handle-item-no-css {{sortable-item model=item groupName="verticle-no-css"}}>
           {{item}}
-          {{#groupItem.handle data-test-vertical-demo-handle-no-css class="handle"}}
-            <span data-item={{item}}>
-              <span>&vArr;</span>
-            </span>
-          {{/groupItem.handle}}
-        {{/group.item}}
+          <span class="handle" data-test-vertical-demo-handle-no-css data-item={{item}} {{sortable-handle}}>
```
MichalBryxi, line 159: Doesn't `{{sortable-handle}}` also need to know the `groupName` as `sortable-item` and `sortable-group`?

MichalBryxi:
In case I forget to add `groupName` to `{{sortable-item}}`, I get following error:

```
sortable-item.js:396 Uncaught TypeError: Cannot read property 'prepare' of undefined
    at SortableItemModifier._startDrag (sortable-item.js:396)
    at SortableItemModifier._prepareDrag (sortable-item.js:361)
    at Backburner._run (backburner.js:1013)
    at Backburner._join (backburner.js:989)
    at Backburner.join (backburner.js:760)
    at join (index.js:168)
    at index.js:263
_startDrag @ sortable-item.js:396
_prepareDrag @ sortable-item.js:361
_run @ backburner.js:1013
_join @ backburner.js:989
join @ backburner.js:760
join @ index.js:168
(anonymous) @ index.js:263
backburner.js:389 Uncaught Error: You attempted to schedule an action in a queue (actions) for a method that doesn't exist
    at DeferredActionQueues.schedule (backburner.js:389)
    at Backburner._scheduleExpiredTimers (backburner.js:1122)
    at Backburner._runExpiredTimers (backburner.js:1095)
```

Maybe asserting against that argument and providing nicely formatted error hinting what should be done would be a good idea?

tests/dummy/app/templates/modifier.hbs
cah-briangantzler: {{sortable-handle}} only really matters when you click on it, so when that happens it looks for the {{sortable-item}} its contained in and gets the group from there.

{{sortable-item}} needs to know some stuff about the group to register it at creation time, before its actually put in the DOM, so group has to be passed to it.      There was another version in which {{sortable-item}} looked up {{sortable-group}} at runtime and you didnt have to pass it in, but there were concerns about performance and hence the service registration came about.

cah-briangantzler:
Good suggestion on the assert

ygongdev:
@cah-briangantzler |

Hey I've been thinking about this problem lately.
So the original intent behind this upgrade was to allow integration with other libraries such as ember-yeti-table, correct?
As I learned more about modifier, I feel like it is not the best solution for us due to several reasons.

1. A modifier is like a pure function. It does not really have the same advantage as contextual components, which can easily manage state. That is why in the modifier approach, we had to use an external resource such as a service to manage state.
2. A service can become problematic since it is a singleton. In this PR, we have to use a unique groupName to identify a sortable-group. Imagine multiple instances of a component using the same sortable-group with the same groupName, I think things would get messy.

What I think we can do is *KEEP* the same architecture, but we convert ember-sortable into template-less/tag-less glimmer components. This will allow you to use custom template such as ember-yeti-table since you are not restricted to any tagname specified by ember-sortable.

Theoretical Integration Example:
```
<EmberSortableGroup ... as |group|>
  <EmberYetiTable ... as |table|>
   <table.header ...>
   
   <group.item>
      <table.column ...>
   </group.item>
  <EmberYetiTable>
<EmberSortableGroup />
```
I haven't tested this yet, but what do you think? Would that address your pain points?

MichalBryxi:
@ygongdev 
I also think that using a service to store the the state feels like a fight against the architecture. And maybe this one will be even more off, but let's put it there:

`EmberSortableGroup` is a good old contextual component. `{{sortable-item}}` is a modifier that consumes respective group component as an argument. This way `sortable-item` is explicitly provided with the context it's supposed to work in:

```hbs
<EmberSortableGroup ... as |group|>
  <table>
    <tr {{sortable-item group=group}}>...</tr>   
  </table>
</EmberSortableGroup
```

I'm not sure about the possibility to build this and usability. I liked the modifier approach at lest for `sortable-item`s as it does not introduce the need to nest stuff under other stuff and gives the code reader nice explicit hint that _this tag_ is the draggable one.

Thoughts?

cah-briangantzler:
@ygongdev this will not work,  sortable needs to move a DOM element, if you make group.item tagless, there is no DOM  element to move. group.item also has no access to the the column as you cant access children of a DOM element that doesnt exist.

@MichalBryxi Opposite problem, you need the container to contain the sortable items. In a table that may be tbody, (so a modifier can be applied) as a wrapper component that it tagless, no access, as a tagged component you have a div, span, whatever in the middle of the table. This is basically why the original component doesnt work. In the docs It does show using the component version in a table, but if you look, the table is not being rendered correctly because of the extra DOM elements. The docs shouldnt show it, it doesnt work.

The original PR did not have a service, did not require a group name  and just worked, but there was concern that on a list of hundreds performance could be poor (this wasnt tested, just a theory). Im also not sure that on a list of hundreds you should be doing drag and drop sorting. This version is the comprise.   Note that modifiers also does not solve a problem should the table be paginated.

ygongdev:
@cah-briangantzler that is true.

How will the modifier approach handle nested sortable-groups? Both this and the other approach.

MichalBryxi:
Ah, I see. So from technical perspective (pseudo code):
```hbs
<GroupDOMElement>
  <ItemDOMEelement />
  <ItemDOMEelement />
</GroupDOMElement>
```
and both types of elements _need_ to be actual DOM nodes so that the user can interact with them.

Component implementation in many cases does not work. It adds that DOM node to interact with, but that one might be breaking HTML structure.

Modifier implementation is in many cases way better, because the user can attach it to any DOM element that is already there.

The problem with modifiers is that individual modifiers have to share state somehow. Hence the service implementation proposed in this PR. And at that point a service for "shared state" does actually sound like a good solution.

So question is: Is there anything _else_ that can hold the state for `sortable-group / sortable-item`? 

Variation on my previously proposed solution. What if encapsulating (tagless) component would hold the state, but the DOM interaction would be done through modifiers?

```hbs
<EmberSortableGroup ... as |group|>
  <table>
    <tbody {{sortable-group onChange=this.update group=group}}>
      <tr {{sortable-item group=group}}>...</tr>   
    </tbody>
  </table>
</EmberSortableGroup>
```

Again, this is just a "braindump" to explore the possibilities.

cah-briangantzler:
I also mulled over this approach and it required the consumer to write HBS for the state component and write `group=group` on the sortable-group modifier and the sortable-item modifier. So I thought, if you have to write `group=group` anyway, thats not far different that `groupName=name` (make this name a property on your component and its basically the exact same statement) but no you no longer need to concern yourself with the additional HBS to create the state wrapper. 

A question was asked about nested groups with this approach, if the group names are different, they nested state is preserved. This is important for a sortable in the parent and a sortable in an outlet (they are both rendered at the same time, but defined in different HBS)

However, thats what this discussion is all about, what consumers of the addon would think as reasonable to do to use the addon.

If a general consensus is that it is preferable to do the additional work in defining the state wrapper as an extra silent component, then that could be done. Perhaps I tried to simplify it too much.

ygongdev:
Given an example like this
`child-component.hbs`

```
{{ember-sortable ... groupName="foo"}}
  {{each ...}}
    <li {{sortable-item ... }} ...> ... </li>
  {{/each}}
```

`parent-component.hbs`
```
{{child-component ...}}
{{child-component ...}}
```

Wouldn't this run into trouble with the groupName approach? What I mean by that is that both `child-component` will be using the same `groupName`, which will create some problems. One workaround is force the consumer to supply a different `groupName` from the `parent-component` level, which is a small amount of additional work.

cah-briangantzler:
yes if you use the same group name. The group name implies something, so I would not use a generic name like foo or group everywhere, that implies nothing. I would use a group name of the component Im in or something more meaningful.

Its been a while since I touched this, so I dont know if the docs have been updated for this yet, but in the docs describing the group name param there would be information to convey, it is used for this state and use meaningful unique names names.

After looking at the code again It should also be noted that groupName is not a required field, there is a default supplied, so for most users, where there are not two sortables on the same screen, you do not have to specify groupName. Just ignore the parameter. The parameter is there ONLY to resolve a conflict when you have two sortables on the screen at the same time. You can choose to give the one in conflict a name, or both.   I wouldnt even bother with group name unless I had a conflict to resolve

@MichalBryxi reviewing the code again for the assert you suggested, I think the problem was not that you forgot to specify a group name on the item, but that you did specify a group name on the group.   Both the group and item have a default as mentioned above, the actual error was not that you forgot one, but that they didnt match.   There does need to be an assert, but not an existence assert like was previously implied, but an assert that the group name specified (the default if you didnt, as in your case) has been registered with the state manager.

MichalBryxi:
@cah-briangantzler 

Ok awesome. Apart from some of the minor fixes, and making sure that CI passes, I can't see any strong objections against merging this?

cah-briangantzler:
I have updated the documentation more around the groupName argument and added the assert as suggested.

The build is currently failing and if you look at the failure it is in the component tests, not the modifier tests. There is no shared code between the component and modifier versions.

If there is something to fix, it should be another PR to the master branch, not included in this PR.

MichalBryxi:
So by using this branch for some experimental implementation I can confirm that the issue that was happening to me and is described in #351 is fixed.

So what is the process now? Are we waiting for some approvals? Are other fixes needed? Shall we first fix the failing unrelated tests?

cah-briangantzler:
I believe I have been added as a reviewer, possibly publisher to this addon, reviewing not a problem, but have not publshed an addon before. When it comes time to release master as a version, if possible I would like to walk through with whoever does it so that should the need arise and i be called upon, I would have at least done it once

ygongdev:
@cah-briangantzler I can share with the steps I use to release on npm. Before that, we have to polish the current `master` a bit more since it'll be most likely a major upgrade `3.x.x`.

cah-briangantzler:
I saw the PR with the updates to doc and migration.  Let me know if theres something you would like me to do

ygongdev:
@cah-briangantzler sure, several tasks:
1. Deleting `components` related code
2. Ensuring test parity on our API and internals (which we might or might not have).
3. Address open issues.

cah-briangantzler:
Since the components and the modifiers share no code, I figured you would deprecate the components, that why people could one by one switch the usage over, and then remove the components in the next release. If you do not have a few releases in which both exist, they would be forced to change all usages in order to go to 3.x

Im fine with the documentation changing to encourage use of the modifier version, but I wouldnt suggest remove the component versions just yet.

ygongdev:
My thought is that since modifier is offering no new features and is a purely a migration to modifier. Keeping them both is going to create bloat as no one should be using both versions in a single app. If the user chooses to stick with components, they would stick with `2.x.x`.

Removing the components and keeping the `modifiers` will show a strict stance that we'll be only supporting the modifier version going forward.

cah-briangantzler:
I agree with all that, except the person that want to convert from component to modifier, they are forced to change them all, rather than change them one at a time. 

The ember way has always been to provide a replacement, deprecate the old, then remove.

So I would suggest doing a 2.x that includes both, mark the components as deprecated, then produce a 3.0 that removes components, much like ember 2.18 contains both ways of doing things, and then 3.0 introduced no new features, just removed the deprecated code

MichalBryxi:
I agree with @cah-briangantzler. DX > few bytes more over the wire.

btecu:
I'd prefer having less bytes over the wire - the version number doesn't matter.

cah-briangantzler:
If you release a 2.2 that contains both with the components deprecated, then a 3.0 with them removed back to back, those that want to convert slowly can go 2.2, those that want to convert all at once or less bytes, can go 3.0, But if someone has like 40 usages of ember-sotrtable in there app, you should force them to do all or nothing. 

So im not saying dont remove the component version, im saying dont do it in the same release as the modifiers are introduced

mike183:
I agree with @cah-briangantzler for the same reasons that he has already stated. 

The Ember philosophy has always been to ensure that upgrade paths are as clear and that applications can be slowly upgraded piece by piece over time.

ygongdev:
@cah-briangantzler we can do that, cut a release with both version and we'll proceed to push only modifier changes into `master`.

PR #365, opened 2020-04-20
Correct typos in modifier examples

danwenzel:
Corrects a few typos in the modifier sections of the README, where `{{group.item}}` was being referenced instead of the `{{sortable-item}}` modifier


ygongdev:
Thanks!

PR #374, opened 2020-06-15
Fixes "'removeChild' of null" in SortableGroupModifier willRemove method

jede:
In some cases the willRemove method is run after the page has navigated away. Therefore we should check if the announcer still has a parentNode before trying to remove it.


ygongdev:
Re-trigger CI

PR #377, opened 2020-07-04
Removed the component version

cah-briangantzler:
Breaking Change !!!

Since 2.2.0 has been released with both component and modifier (Although I would have liked to see the component version deprecated instead of just mentioned in readme) this PR removes the component version.  It does not update the readme as there is another PR pending that does that


ygongdev:
Looks like tests are failing. Since we only have modifiers in this PR, we should look into stabilizing the tests.

cah-briangantzler:
I removed the component integration tests, looks like I missed the component acceptance tests. Sorry, will fix that

cah-briangantzler:
So whats the difference between the tests and the one with floating dependancies? I dont know why tests would pass and floating would fail one.

ygongdev:
I'm actually not sure but it looks like the floating dependencies are missing the `yarn install` and `yarn lint` step. I wonder if that means something like `yarn.lock` needs to be updated?

cah-briangantzler:
running yarn install did not update the yarn.lock

ygongdev:
Mmm, I dug into the past failures. It could be a flaky or legit test failure for `Acceptance | smoke modifier: reordering with mouse events` for Firefox.

ygongdev:
@jgwhite Sorry to bother you, but do you remember what the intention behind the  "Floating dependencies" test is meant to test for?

jgwhite:
> @jgwhite Sorry to bother you, but do you remember what the intention behind the  "Floating dependencies" test is meant to test for?

I have no idea. Sorry!

cah-briangantzler:
Looking into it more it looks like it runs the tests without the yarn lock. To see if it runs with the most recent versions.

However, for an addon, there is no reason to use a yarn lock. When the addon is consumed in the application, its the applications yarn.lock that applies, the one in the addon is ignored isnt it?

That being said, let me try running locally with no yarn lock and see what I get.

cah-briangantzler:
This https://github.com/adopted-ember-addons/ember-sortable/pull/380 should be vetted and merged and another release before merging in removing components

cah-briangantzler:
Ok so without a yarn.lock I get no failures when running `ember t -s` however do get the one failure when running `ember t`.  I notice that without the yarn.lock we are getting all the errors `WARN: DEPRECATION: setSourceDestroying is deprecated, use the destroy() API to destroy the object directly` which you can refer to https://github.com/kaliber5/ember-bootstrap/pull/1143

snewcomer:
@cah-briangantzler Mind merging master?  We are close to ready for a major version bump

cah-briangantzler:
rebased, push, rebuilding now

snewcomer:
@cah-briangantzler This looks great!  Mind updating the README as well?

cah-briangantzler:
Updated.

jacojoubert (approved):

snewcomer (approved):
ref https://github.com/adopted-ember-addons/ember-sortable/commit/ffbd488b745f41ba4077e7e8d37410b247e89435

kiwiupover:
We are not going to release the new version until we have the tests passing on ember 3.20, 3.24, and 3.28 PRs for those changes will follow. #450

kiwiupover:
@cah-briangantzler thanks heaps for pushing this work forward.

bjornharrtell:
This was not noted as a breaking change at https://github.com/adopted-ember-addons/ember-sortable/blob/master/CHANGELOG.md#v300-2021-11-29, I guess it should be?

PR #380, opened 2020-07-28
Fixed modifier tracking correct dropped item with keyboard events

cah-briangantzler:
Fixes https://github.com/adopted-ember-addons/ember-sortable/issues/379


ygongdev:
re-trigger ci

PR #388, opened 2020-09-24
upgrade ember-test-selectors to v4

nlfurniss:
This upgrades ember-cli-babel to v7. Also removed `ember-native-class-polyfill` since it's not needed anymore


ygongdev:
re-triggering checks

ygongdev:
Can we fix the yarn.lock?

nlfurniss:
> Can we fix the yarn.lock?

Done!

PR #400, opened 2020-11-12
fix(deprecation): update ember-test-selectors

chrismllr:
Resolves getWithDefault deprecation.

**Release Notes for 5.0.0**
https://github.com/simplabs/ember-test-selectors/blob/master/CHANGELOG.md#v500-2020-09-02

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "ember-cli-htmlbars": "^5.0.0",
     "ember-decorators-polyfill": "^1.1.1",
     "ember-modifier": "^2.1.0",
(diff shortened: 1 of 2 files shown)


chrismllr:
Looks like the tests will need to be refactored for the 3.20 scenario to pass; still using `moduleForComponent` for component tests

cah-briangantzler:
Is there still a plan to remove the component version? If that is the case those tests would be deleted as well, no need to fix

ygongdev:
I don't think the failure is related to moduleForComponent. Should be safe to merge

Issue #406, opened 2021-01-19
ember-get-config should be a dependency


smfoote:
**Describe the bug**
`ember-get-config` is used in the `sortable-item` modifier, but is listed as a `devDependency` instead of a production `dependency`. As such, it is excluded from production builds. Any consumer that is using `sortable-item` in production will break in production.

**To Reproduce**
1. Create a production build
2. Note that `ember-get-config` is not included in the bundle.

**Expected behavior**
`ember-get-config` should be included in the production bundle.

PR #407, opened 2021-01-19
Move `ember-get-config` to `dependencies`

smfoote:
Addresses #406


ygongdev (commented):
Quick sanity check: did you ran `yarn install` after this change?
Want to see if there should be any necessary `yarn.lock` changes

smfoote:
Yes, I ran `yarn install`. I think there's no change because it's already installed, just moved to `dependencies`.

nbibler:
@smfoote and @ygongdev : Thank you! I came here to report this issue and found that you'd identified it, patched it, and it'd already been released! 👏

PR #411, opened 2021-01-26
Allowed waiting on an optional promise before resetting items

kobsy:
### Summary
This is a fantastic addon and a slick implementation of drag-and-drop sorting! However, as my team and I were using it, we found we wanted to tie the "drop" action to a network request to persist the new order to the server. While straightforward when using an ORM like Ember Data (just update position, which updates the view, and then call `save()`), it's more complicated if the underlying data is immutable. Ideally, when using something like Apollo to communicate with a GraphQL API, you'd be able to "hold" the rearranged items in-place in the UI while the mutation is sent, finally updating the data powering the view (automatically when Apollo updates its cache) after the request completes.

This PR is an attempt to address this use-case by means of an optional promise returned from the `onChange` action. If the action returns a promise, then we wait for it to either resolve or reject before we schedule the reset of the items' positions as managed by `ember-sortable`.

For existing usage, there should be effectively no change. Although the order of calling `onChange()` and `_updateItems()` is reversed at their respective call-sites, because `_updateItems()` schedules its UI work using the Ember runloop, nothing should get out-of-sync, as `onChange` would continue to be run as part of the `action` queue. For the case where `onChange` returns a promise, however, `_updateItems()` would now be called as part of its `finally()` callback, only resetting the items' positioning/CSS after the promise either rejects or resolves. My hope was that the change could truly be additive, not interfering with the way the addon functions for the vast majority of users.

All the tests that currently pass continue to pass with these changes -- I wasn't able to track down the failures on Ember 3.20, but they seem to be happening even on the latest release -- but neither was I quite sure how to test the new functionality. Since the DOM itself isn't rearranged until after the underlying data is modified, it's not straightforward to check that the rearrangement is "holding" until the promise resolves. If there's a way that would make sense to be able to check this, I'd be happy to add a test case or two using that method. 🙂

Thanks for your consideration!

### DIFF
--- a/addon/components/sortable-group.js
+++ b/addon/components/sortable-group.js
@@ -267,12 +267,20 @@ export default Component.extend({
     this.set('moves', []);
     this._disableKeyboardReorderMode();
     this._tearDownA11yApplicationContainer();
-    this._updateItems();
+
+    let promise;
     if (groupModel !== NO_MODEL) {
-      this.onChange(groupModel, itemModels, selectedModel)
+      promise = this.onChange(groupModel, itemModels, selectedModel)
     } else {
-      this.onChange(itemModels, selectedModel);
+      promise = this.onChange(itemModels, selectedModel);
     }
+
+    if (promise && typeof promise.finally === 'function') {
+      promise.finally(() => this._updateItems());
+    } else {
+      this._updateItems();
+    }
+
     this._updateItemVisualIndicators(_selectedItem, false);
     this._updateHandleVisualIndicators(_selectedItem, false);
     this._resetItemSelection();
@@ -380,11 +388,17 @@ export default Component.extend({
       draggedModel = get(draggedItem, 'model');
     }
 
-    this._updateItems();
+    let promise;
     if (groupModel !== NO_MODEL) {
-      this.onChange(groupModel, itemModels, draggedModel);
+      promise = this.onChange(groupModel, itemModels, draggedModel);
     } else {
-      this.onChange(itemModels, draggedModel);
+      promise = this.onChange(itemModels, draggedModel);
+    }
+
+    if (promise && typeof promise.finally === 'function') {
+      promise.finally(() => this._updateItems());
+    } else {
+      this._updateItems();
     }
   },
 
--- a/addon/modifiers/sortable-group.js
+++ b/addon/modifiers/sortable-group.js
@@ -652,15 +652,17 @@ export default class SortableGroupModifier extends Modifier {
       draggedModel = draggedItem.model;
     }
 
-    this._updateItems();
-    this._onChange(itemModels, draggedModel);
+    const promise = this._onChange(itemModels, draggedModel);
+    if (promise && typeof promise.finally === 'function') {
+      promise.finally(() => this._updateItems);
+    } else {
+      this._updateItems();
+    }
   }
 
   @action
   _onChange(itemModels, draggedModel) {
-    if (this.onChange) {
-      this.onChange(itemModels, draggedModel);
-    }
+    return this.onChange && this.onChange(itemModels, draggedModel);
   }
 
   /**
--- a/tests/acceptance/smoke-component-test.js
+++ b/tests/acceptance/smoke-component-test.js
@@ -1,5 +1,5 @@
 import { module, test } from 'qunit';
-import { visit, find, findAll, triggerKeyEvent, focus, blur } from '@ember/test-helpers';
+import { visit, find, findAll, triggerKeyEvent, focus, blur, waitFor } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import { drag, reorder }  from 'ember-sortable/test-support/helpers';
 import { ENTER_KEY_CODE, SPACE_KEY_CODE, ESCAPE_KEY_CODE, ARROW_KEY_CODES } from "ember-sortable/test-support/utils/keyboard";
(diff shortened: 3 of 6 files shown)


ygongdev:
Thanks for the nice feature PR!

I just have some questions:
1. It seems like by using `.finally`, we're forcing the user to accept the UI update whether or not the promise resolves or rejects. There might be use cases where we want `_updateItems` only when the promise resolves, not when it rejects. So it might be a nice to think of some configurability around that.
2. As for testing, yeah it can be a bit tricky since we're manipulating the DOM order on the fly when reordering and the tests might not necessarily catch the correct order at the time we assert. However, I think if we can add a few smoke scenarios on the demo page, that would be better than nothing.

kobsy:
Great questions!

1. I initially was using `then` but ended up changing it to `finally`, and my reasoning was that I didn't want the user to be able to get into a state where the items _never_ reset after a drag. In a way, having the reset happen in `finally` closely reflects what the addon was doing already; rather than having the items repositioned by the addon indefinitely, after the promise has completed, regardless of the outcome, they go back to being displayed "as they are" in the DOM after completion (whether with updated data and a new layout or with the old data and the original layout). I wouldn't be opposed to finding a way to offer more flexibility, but I'm not sure how to do that without making the API more complex or allowing the user to get the addon into an inconsistent state with items having never reset. 😓 
2. I'll go ahead and add a scenario to the smoke test on the demo page, then. 🙂

ygongdev:
> 1. inconsistent state with items having never reset

By resetting, do you mean updating the UI, e.g `_updateItems`?
I think there might be scenario where the API fails, we don't want to update the UI right? We would want the original state, so we would never trigger `_updateItems` and the other corresponding UI updates. 
Unless you're thinking about some other inconsistent state(s)?

kobsy:
I think we might have different ideas of what should happen if the promise should reject/the API fail (which would support your suggestion that `_updateItems` be tied to `then()` instead of `finally()`), but I could also just be misunderstanding what all `_udpateItems` is doing.

I picture the situation like this:
You have a list automatically sorted by a `position`:
1️⃣ Apples
2️⃣ Pears
3️⃣ Oranges

Then you drag to rearrange:
3️⃣ Oranges
1️⃣ Apples
2️⃣ Pears

At this point, the UI reflects the drag. You send off your request. If it succeeds, then the underlying positions are updated, and `_updateItems` removes the UI customizations and Ember re-renders the view to reflect the new order:
1️⃣ Oranges
2️⃣ Apples
3️⃣ Pears

If the API fails and the promise rejects, what should happen? I was thinking that you'd revert the list back to its pre-drag state by calling `_updateItems` with unmodified data:
1️⃣ Apples
2️⃣ Pears
3️⃣ Oranges

If, however, what you wanted was to leave the sort in the new order (perhaps in hope that a subsequent drag's request would go through and persist _that_ new state) even though it wasn't persisted on the backend, then I guess you wouldn't have to call `_updateItems` at all, and the state of the UI would be out-of-sync with that of the backend (at this point, maybe you'd have the option of prompting the user to revert or retry?). This is the situation I was trying to avoid by calling `finally()` instead of `then()`.

I can certainly change it to use `then()`; this would put the onus on the user -- if they wanted to avoid this last situation -- to have their promise always resolve, perhaps by wrapping the API call in another promise which could gracefully handle API errors but always successfully resolve as far as `ember-sortable` is concerned. 🤷 If this is the best way forward for the addon, I'm happy to make the change on the PR.

On the other hand, if the promise rejects and `_updateItems` is never called, there's not a good hook to reset that UI state (revert the drag, e.g.) -- `_updateItems` will end up never being called unless the user attempts another drag (and _that_ promise resolves). I think for me, I'd rather have the addon work largely as it does now -- where `_updateItems` is called after _every_ drag, regardless -- but just allowing for that promise-based delay that allows a bit more time for updating the data.

In any case, please advise on how you want to move forward with this PR, and I'll make whatever changes are requested. Thanks again! 🙂

cah-briangantzler:
If I recall this is a Data Down Actions Up problem. The data is passed to ember sortable which displays it, the user drags an item and the action is called informing the caller of the new desired order. You will notice that example onChange is
```
@action
  update(newOrder, draggedModel) {
    set(this, 'model.items', newOrder);
    set(this, 'model.dragged', draggedModel);
  }
```
You know what happens if you delete the line 
```
      set(this, 'model.items', newOrder);
```

The drag never happens. 

Why this is important is that it is not ember sortable reordering the data. It is the application setting the new order.   If the application needs this kinda of behaviour, the application should save the original state, do the set so ember-sortable displays the move, execute the promise in the onChange action and if it rejects, set the data back to the saved state.

Logic of saving a state, then waiting to reset the state is not the concern of the component. It would be very similar to changing the input field onChange event (fired when it loses focus) to allow for a promise, and if the promise fails the input field would be responsible for reseting the data back to its initial value.   We dont do that, we have mechanics in place to already handle this. It is the applications responsibility to reset the value passed into the input field if the promise fails.

I do not believe this kinda of logic is within the scope of this components concern. 


As an aside for the application code, have you considered what if it take 3 seconds for the promise to complete, they could have then performed another few drags, so now you got 3 promises pending and if the first one rejects, what order do you set it back to, do you undo the other two pending drags, what if they succeed?  Or are you going to lock the component for no future drags until the promise completes.

ygongdev:
All good points here. 
So` _updateItems` is actually responsible for the reflecting the final UI reordered state because it syncs the internal sorted list with the DOM.
@cah-briangantzler brings up a good point regarding multiple promises. If we handle promise internally in ember-sortable, the UI could also very easily go out of sync if the promises do not resolve in the order they come in.
Even if the promise goes through, if it takes a long time and `_updateItems` is not triggered in time, it might also create temporarily visual defect.
Will need to think on this problem a bit more.

kobsy:
👍 Got it. Thanks for the feedback, both of you! I'll go ahead and close this PR for now. While I still like the idea of the addon waiting for the consuming component to finish before resetting its internal state, I also recognize how it does bend the DDAU principle a bit, 😅 as well as how (from trying to write tests for the new case) it adds more complexity than might justify the change, especially when doing several drags in a row. We'll work on this from the application side for now. Again, thanks for the feedback, as well as for the incredible addon and all the work you do maintaining it! 🙂

cah-briangantzler:
I would picture you doing something in the action in the application in the onChange. Once you have that code working, if you extract it to a reusable state, I could see that as a helper addon or a helper util in this addon that consumers could opt to use. 

That way you would have much more control over configuring the needs of this helper (additional actions maybe) rather than adding additional config to the ember-sortable interface.

PR #424, opened 2021-06-08
Set default values for itemVisualClass  and a11yAnnouncementConfig

jacojoubert:
This adds reasonable defaults for `itemVisualClass` and `a11yAnnouncementConfig` so sortable can be used with less customization required on each invocation. 

There as already default values for `is-dragging` and `is-dropping` used by mouse re-ordering, so it made sense to follow that pattern and use `is-activated` for the keyboard version. 

The `a11yAnnouncementConfig` is copied from the example docs.

### DIFF
--- a/addon/utils/defaults.js
+++ b/addon/utils/defaults.js
@@ -1,0 +1,25 @@
+export const defaultA11yAnnouncementConfig = {
+  ACTIVATE({ a11yItemName, index, maxLength, direction }) {
+    let message = `${a11yItemName} at position, ${index + 1} of ${maxLength}, is activated to be repositioned.`;
+
+    if (direction === 'y') {
+      message += 'Press up and down keys to change position,';
+    }
+    else {
+      message += 'Press left and right keys to change position,';
+    }
+
+    message += ' Space to confirm new position, Escape to cancel.';
+
+    return message;
+  },
+  MOVE({ a11yItemName, index, maxLength, delta }) {
+    return `${a11yItemName} is moved to position, ${index + 1 + delta} of ${maxLength}. Press Space to confirm new position, Escape to cancel.`;
+  },
+  CONFIRM({ a11yItemName }) {
+    return `${a11yItemName} is successfully repositioned.`;
+  },
+  CANCEL({ a11yItemName }) {
+    return `Cancelling ${a11yItemName} repositioning`;
+  },
+};
--- a/README.md
+++ b/README.md
@@ -364,7 +364,7 @@ There are 4 modes during keyboard navigation:
 #### User configurable
 ##### Screen Reader
 - **a11yItemName**
-  a name for the item.
+  a name for the item. Defaults to `item`.
 - **a11yAnnouncementConfig**
   a map of `action enums` to `functions` that takes the following `config`, which is exposed by `sortable-group`.
 ```javascript
@@ -378,7 +378,7 @@ a11yAnnounceConfig = {
 ```
 and returns a `string` constructed from the `config`.
 
-**Example**
+**Default value**
 ```javascript
 {
   ACTIVATE: function({ a11yItemName, index, maxLength, direction }) {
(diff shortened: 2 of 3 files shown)


ygongdev:
Can you also remove the `NO_MODEL` check [here](https://github.com/adopted-ember-addons/ember-sortable/blob/master/addon/components/sortable-group.js#L549)?

I have another request that is a bit unrelated to your PR, but I think would be great to mention somewhere in the docs that when you create your `a11y configurations`, you can also use any i18n tools to make the a11y message more i18n friendly.

jacojoubert:
I made the changes.

In the future Auditboard will be localizing their application. What do you think about moving the announcer behavior to the ember-sortable service and allowing the consuming app to extend and override it? This will enable the global a11y customization and solve the problem of having to pass in the config on every invocation.

kiwiupover:
@ygongdev this is a great PR too

ygongdev:
> solve the problem of having to pass in the config on every invocation.

I don't think you would want that if your application is using the sortable under a different context right? In fact, allowing you to use different config on every invocation should be an advantage because you can control the messaging per instance.

If you really want global a11y customization, you can also just make a wrapper component around `ember-sortable` that uses your preferred a11y messaging?

We can take this discussion as another issue/PR since it should be unrelated to the current PR.

addon/utils/defaults.js
```diff
@@ -0,0 +1,25 @@
+export const defaultA11yAnnouncementConfig = {
```
MelSumner, line 1: Could you add a test for this?

cah-briangantzler:
What you may want to do is add it to the service as a default. That way if you do want the same config on all your uses, or at least the majority, you dont have to keep passing it in. The you would only pass it when you needed to change it.

jacojoubert:
> What you may want to do is add it to the service as a default. That way if you do want the same config on all your uses, or at least the majority, you dont have to keep passing it in. The you would only pass it when you needed to change it.

Yeah, that is what I was thinking.

addon/utils/defaults.js
jacojoubert: done

MelSumner (approved):
Looks good. Let's get @ygongdev to take one more look before merging. Thank you for doing this work! 👍

tests/integration/modifiers/sortable-group-test.js
```diff
@@ -54,6 +54,49 @@ module('Integration | Modifier | sortable-group', function(hooks) {
...
+      </ol>
+    `);
+
+
+    triggerKeyEvent('[data-test-handle=Uno]', 'keydown', 32) /* SPACE */;
+
+    await announcerHasText();
+    assert.dom(announcerSelector).hasText('item at position, 1 of 3, is activated to be repositioned.Press up and down keys to change position, Space to confirm new position, Escape to cancel.');
+
+    triggerKeyEvent('[data-test-handle=Uno]', 'keydown', 40) /* DOWN */;
```
ygongdev, line 83: Should we await `triggerKeyEvent`? Does`waitUntil` from `await announcerHasText` properly `await` for settleness?

ygongdev:
> > What you may want to do is add it to the service as a default. That way if you do want the same config on all your uses, or at least the majority, you dont have to keep passing it in. The you would only pass it when you needed to change it.
> 
> Yeah, that is what I was thinking.

Feel free to start a new issue to discuss this more thoroughly. :)

tests/integration/modifiers/sortable-group-test.js
jacojoubert: I can't await settledness, because that would be too late to catch the announcement text. The announcement text gets removed from the dom pretty quickly. The `await announcerHasText` basically polls the dom for changes to the announcer text element and as soon as it changes it resolves.

ygongdev: gotcha.

ygongdev (approved):
LGTM

PR #425, opened 2021-06-09
Require ember-decorators-polyfill to be installed by client app

### LINKED ISSUE #378: Ember-decorators-polyfill error after upgrading to 2.2.0
**Describe the bug**
After upgrading to latest version I'm getting this error on build:

`Cannot destructure property 'attr' of 'DS' as it is undefined.`

Which stems from `data-fix.js` a file contained in the `ember-decorators-polyfill` dependency.

This issue was raised on the polyfill repo [here](https://github.com/ember-polyfills/ember-decorators-polyfill/issues/29) and looks to have only been a problem with version `^1.1.1`  however I've tried using the latest version (`1.1.5`) locally and I get the same error.

I'm not sure if the issue lies here or with the polyfill so I thought I'd raise it here first, as I'd imagine we should at least up the polyfill version to the latest.

Happy to raise a PR to up the version if anyone agrees.

Another option could be to make this polyfill optional like `closest`, as we shouldn't actually need it over Ember 3.10, but that would be a bigger job I imagine.

**To Reproduce**
1. Install version 2.2.0 of this package
2. Run `ember serve`

**Screenshots**
Screenshot 2020-07-14 at 10 10 38

**Additional context**
We're on Ember `3.10.2` if that helps.

jacojoubert:
This removes the ember-decorator-polyfill dependency and instead adds a console warning if the consuming app doesn't have it installed. The issue here is that the polyfill breaks older versions of ember-data when used with the newest versions of ember-source. For users stuck on old versions of ember-data due to performance regressions this can be an issue. 

Fixes #378

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ For modifiers, please read this [migration guide](/MIGRATION_GUIDE_MODIFIERS.md)
 
 In version 2.0.0+, our `closest` polyfill seems to break some app's `production` build. To mitigate this, the `closest` polyfill will only enabled if it doesn't break the `production` build (if the `polyfill` file is recognized by the build). Affected apps will need to supply their own [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill) polyfill to ensure compatibility with IE. This issue is tracked [here](https://github.com/adopted-ember-addons/ember-sortable/issues/333).
 
+We require yhe `ember-decorators-polyfill` to be installed if you are using this addon with versions of Ember < 3.10.0.
+
 Version 1.0 depends upon the availability of 2D CSS transforms.
 Check [the matrix on caniuse.com](http://caniuse.com/#feat=transforms2d)
 to see if your target browsers are compatible.
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ For modifiers, please read this [migration guide](/MIGRATION_GUIDE_MODIFIERS.md)
 
 In version 2.0.0+, our `closest` polyfill seems to break some app's `production` build. To mitigate this, the `closest` polyfill will only enabled if it doesn't break the `production` build (if the `polyfill` file is recognized by the build). Affected apps will need to supply their own [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill) polyfill to ensure compatibility with IE. This issue is tracked [here](https://github.com/adopted-ember-addons/ember-sortable/issues/333).
 
-We require yhe `ember-decorators-polyfill` to be installed if you are using this addon with versions of Ember < 3.10.0.
+We require the `ember-decorators-polyfill` to be installed if you are using this addon with versions of Ember < 3.10.0.
 
 Version 1.0 depends upon the availability of 2D CSS transforms.
 Check [the matrix on caniuse.com](http://caniuse.com/#feat=transforms2d)
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -15,7 +15,8 @@ module.exports = function() {
           name: 'ember-lts-3.8',
           npm: {
             devDependencies: {
-              'ember-source': '~3.8.0'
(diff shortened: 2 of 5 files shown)


kiwiupover:
@ygongdev This PR fixes the issues I was running into this morning. @joshuaswift does this PR fix your issues?

ygongdev:
> The issue here is that the polyfill breaks older versions of ember-data when used with the newest versions of ember-source. For users stuck on old versions of ember-data due to performance regressions this can be an issue.

Do you mind giving some more context around this issue? Just curious to know what the root cause is before moving ahead with PR (which I would suggest adding to the compatibility section of the README).

cah-briangantzler:
I believe you are suppose to list this as a peerDependancy the way ember-quint for qunit https://github.com/emberjs/ember-qunit/blob/master/package.json#L70 if you look into the docs for the upgrade to version 5 they talk about this.

jacojoubert:
The issue is that `ember-decorator-polyfill` also polyfills ember-data, and we don't want it to. This causes an `Cannot destructure property 'attr' of 'DS' as it is undefined.` error. Why I am not sure, I didn't go down that afar into the rabbit hole. We are on version 3.4.4 of ember-data and unable to upgrade due to severe performance regressions.

jacojoubert:
> I believe you are suppose to list this as a peerDependancy the way ember-quint for qunit https://github.com/emberjs/ember-qunit/blob/master/package.json#L70 if you look into the docs for the upgrade to version 5 they talk about this.

I am not sure its appropriate here. It's not a real dependency, it's only a dependency if your version of ember-source is out of date. If you are running the latest ember you would get an unmet peer dependency error which is not ideal.

README.md
```diff
@@ -24,6 +24,8 @@ For modifiers, please read this [migration guide](/MIGRATION_GUIDE_MODIFIERS.md)
 
 In version 2.0.0+, our `closest` polyfill seems to break some app's `production` build. To mitigate this, the `closest` polyfill will only enabled if it doesn't break the `production` build (if the `polyfill` file is recognized by the build). Affected apps will need to supply their own [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#Polyfill) polyfill to ensure compatibility with IE. This issue is tracked [here](https://github.com/adopted-ember-addons/ember-sortable/issues/333).
 
+We require yhe `ember-decorators-polyfill` to be installed if you are using this addon with versions of Ember < 3.10.0.
```
jaredgalanis, line 27:
```suggestion
We require the `ember-decorators-polyfill` to be installed if you are using this addon with versions of Ember < 3.10.0.
```

jaredgalanis (approved):
LGTM 👍

rwwagner90 (approved):

ygongdev (approved):

kellyselden:
This is a breaking change, right? For users on old ember without ember-decorator-polyfill already, they will break updating to this? This was released in the latest patch release. Should it be reverted?

kiwiupover:
This is already broken for teams on older EmberData versions  

On July 15, 2021, GitHub ***@***.***> wrote:
> This is a breaking change, right? For users on old ember without
> ember-decorator-polyfill already, they will break updating to this?
> This was released in the latest patch release. Should it be reverted?
>

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/adopted-ember-addons/ember-
sortable/pull/425#issuecomment-880844587>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAE6EC7AIMYXN6HJKNBTTCTTX4ET5ANCNFSM46MS2SFQ>.

kellyselden:
@kiwiupover I think this is more than an ember-data issue.

> We require the ember-decorators-polyfill to be installed if you are using this addon with versions of Ember < 3.10.0.

So for Ember 3.9 users that try to update from v2.2.4 to v2.2.5, they will break since they were relying on ember-decorators-polyfill coming from this addon.

PR #449, opened 2021-11-09
relax `@ember/render-modifiers` dependency

miguelcobain:
This should allow people to update their ember apps more easily.


package.json
```diff
@@ -24,7 +24,7 @@
     "release": "release-it"
   },
   "dependencies": {
-    "@ember/render-modifiers": "^1.0.2",
+    "@ember/render-modifiers": "^1.0.2 || ^2.0.0",
```
snewcomer, line 27: Does updating the lockfile do anything?  Might need it

miguelcobain: The lockfile of a dependency isn't really used be the consuming app, AFAIK.
It could affect the dummy app, though, but

miguelcobain: In any case, I updated the lockfile.

snewcomer: Mind rebasing with master and we will cut with 4.0?

miguelcobain: @snewcomer done

kiwiupover:
@miguelcobain Thanks you so much for this PR. I'm glad we can finally get it merged.

PR #452, opened 2021-11-26
General updates to dependencies

snewcomer:


kiwiupover (approved):

.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = {
+  singleQuote: true,
```
kiwiupover, line 4: Let's add `printWidth:  120` too.

PR #456, opened 2021-12-31
Upgrade `release-it` packages

alexlafroscia:
I've been having authentication issues with `npm`, and figured that getting the `release-it` packages upgraded to their latest versions might help.

From what I can tell... it did not fix my problems. But upgrading these packages doesn't hurt anyway!


kiwiupover (approved):

MelSumner (approved):
Thanks for doing this work!

PR #468, opened 2022-03-13
Fix README variable assignment syntax

mrloop:


README.md
```diff
@@ -65,8 +65,8 @@ The `onChange` action is called with two arguments:
 export default class MyRoute extends Route {
   @action
   reorderItems(itemModels, draggedModel) {
```
NullVoxPopuli, line 67: would you also be willing to update the example? 
 - currentModel is ???
 - the template above references `this.reorderItems`, yet this is on a route (we should probably change everything to be in the form of a component demo / example

README.md
```diff
@@ -60,26 +62,29 @@ The `onChange` action is called with two arguments:
 - The model you just dragged
 
 ```js
-// app/routes/my-route.js
+// app/components/my-list/component.js
+
+export default class MyList extends Component {
+  @tracked lastDragged;
+  @tracked items = ['Coal Ila', 'Askaig', 'Bowmore'];
```
mrloop, line 69: Does this over simplify it, should I use object literals instead so can do `{{item.name}}` as in current examples?

NullVoxPopuli: nah, I think this is good :100:

NullVoxPopuli:
thanks for doing this!

PR #476, opened 2022-05-06
Update ember-modifier to 3.2.0

BnitoBzh:
Fix https://github.com/adopted-ember-addons/ember-sortable/issues/474


BnitoBzh:
someone to confirm? I don't think there is a breaking change.

cah-briangantzler:
Oh he made it opt in to Lazy, thought it was lazy by default.  https://github.com/ember-modifier/ember-modifier/blob/master/MIGRATIONS.md#when-is-this-a-breaking-change

So yea, prob not a breaking change

NullVoxPopuli:
looks good to me -- much easier to read

PR #477, opened 2022-05-16
perf: avoid creating callbacks for the destructor

BnitoBzh:
Extract `cleanup` to a function in module scope which accepts an
instance as its public argument, and use that when calling
`registerDestructor` as well as at the start of every `modify` call.
This results in having exactly one function allocated for the
destructor to call, instead of a function per instance. This is a
relatively small win, but a meaningful one which might add up
significantly if you have many of these on a given page.


NullVoxPopuli:
thank you!!!

PR #485, opened 2022-08-29
when item is disabled, do not add event listeners

st-h:
This PR makes sure event listeners on sortable-item are only added when the item is not disabled, as it appears to be a waste of resources to add event listeners to an element where they are not needed. Sortable group already does not register listeners when in disabled state. 

Background: In a template, I do not want to maintain a list for display purposes and a separate list for editing. Therefore I would like to have a list that where I can enable editing, that does not consume unnecessary resources in display state.

This PR makes #353 obsolete.

### DIFF
--- a/addon/modifiers/sortable-item.js
+++ b/addon/modifiers/sortable-item.js
@@ -849,7 +849,7 @@ export default class SortableItemModifier extends Modifier {
 
     if (this.named.disabled && this.listenersRegistered) {
       this.removeEventListener();
-    } else if (!this.args.named.disabled && !this.listenersRegistered) {
+    } else if (!this.named.disabled && !this.listenersRegistered) {
       this.addEventListener();
     }
   }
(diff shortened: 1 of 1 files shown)


addon/modifiers/sortable-item.js
```diff
@@ -835,11 +842,16 @@ export default class SortableItemModifier extends Modifier {
     }
 
     if (!this.didSetup) {
-      this.addEventListener();
       this.element.dataset.sortableItem = true;
       this.sortableService.registerItem(this.groupName, this);
       this.didSetup = true;
     }
+
+    if (this.args.named.disabled && this.listenersRegistered) {
```
bertdeblock, line 850: If you rebase, you can do `this.named.disabled` as using `this.args` is deprecated.

st-h:
@bertdeblock updated

knownasilya (approved):

PR #488, opened 2022-09-01
Remove unused dependencies

bertdeblock:
- `@ember/render-modifiers` isn't needed as this addon doesn't use any render modifiers anymore
- `ember-cli-htmlbars` can be moved to `devDependencies` as this addon doesn't ship any templates anymore
- `ember-cli-version-checker` isn't used anymore
- `ember-test-selectors` can be moved to `devDependencies` as this addon doesn't ship any templates anymore containing `data-test-` selectors (the dummy app does have `data-test` selectors)

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
   "dependencies": {
     "@ember/test-waiters": "^3.0.1",
     "ember-cli-babel": "^7.26.11",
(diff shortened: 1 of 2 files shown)


gilest (approved):
Nice improvements. All makes sense to me

Windvis (approved):

PR #505, opened 2022-11-23
SLS- 1218: Touch click after dropping item does not work

BvG-Gynzy:


BvG-Gynzy:
Created in incorrect fork

PR #511, opened 2023-01-28
Prep for Release (v5, Major)

NullVoxPopuli:
Just cleaning up a few things after some PRs were merged, prepping for successful release! :tada: 

What's included?
 - fix release-it config
 - update rollup config
 - change some dependencies to peer dependencies to reduce conflicts with app's versions, see: https://github.com/adopted-ember-addons/ember-sortable/pull/512
   - because yarn (v1) can't handle peers, we now need to convert the monorepo to pnpm, which can handle peers
 - update README to note that v2 addons have no node requirement / restrictions.  
 - move test-support utilities to be from a single export to improve discoverability and reduce cognitive load -- this will doubly benefit during a typescript conversion in a future minor release
 - drop support for ember-classic (only Octane supported)


.github/actions/assert-build/action.yml
```diff
@@ -0,0 +1,30 @@
...
+  steps:
+  - name: Build and Assert Output
+    shell: bash
+    run: |-
+      echo '
+        target: ./addon/dist
+        setup:
+          run: pnpm build
+          cwd: ./addon/
+        expect: |
```
NullVoxPopuli, line 14: this mostly ensures:
 - build succeeds
 - sourcemaps are emitted
 
 when TS is added, we also test:
  - if declarations are added
  - if declaration maps are added

.github/actions/download-built-package/action.yml
```diff
@@ -0,0 +1,15 @@
+name: Download built package from cache
+description: Download built package from cache
+runs:
+  using: "composite"
+  steps:
+  - name: Download built package from cache
```
NullVoxPopuli, line 6: re-use the built assets from the first C.I. job so we reduce variability between jobs, and ensure we work with the same assets throughout the C.I. run -- this is as close to simulating a package release (without releasing), as I could come up with

.github/actions/setup-job/action.yml
```diff
@@ -0,0 +1,14 @@
+name: Setup node and package manager
```
NullVoxPopuli, line 1: extract boilerplate so that the main ci.yml is easier to read

.github/workflows/ci.yml
```diff
@@ -12,86 +12,74 @@ concurrency:
...
-      working-directory: addon
-    - name: Lint Test App
-      run: yarn lint
-      working-directory: test-app
-    - name: Lint Docs
-      run: yarn lint
-      working-directory: docs
+    - uses: ./.github/actions/setup-job
+    - name: Lint
+      run: pnpm lint
```
NullVoxPopuli, line 32: top-level lint lints all workspaces at once, in parallel

README.md
```diff
@@ -18,7 +18,9 @@ Sortable UI primitives for Ember.
 |----------------|-------|------|
 | 3.0.0          | 3.24+ | 12+  |
 | 4.0.0          | 3.24+ | 14+  |
-| 5.0.0          | 3.28+ | 14+  |
+| 5.0.0          | 3.28+ | n/a[^1]  |
+
+[^1]: Node is not relevant for v2 addons. As of v5.0.0, ember-sortable is a v2 addon. V2 addons don't have to be for browser-only contexts, but ember-sortable is -- so node is not relevant anymore. This is different from v1 addons, which are not browser libraries, but node programs that happen emit browser code to the consuming app at app-build time.
```
NullVoxPopuli, line 23: This is an important distinction for v2 addons that may be a bit different from what folks are used to

README.md
```diff
@@ -343,20 +345,20 @@ and returns a `string` constructed from the `config`.
...
-import { drag, reorder } from 'ember-sortable/test-support/helpers';
-import {
+import { 
+  drag, reorder,
   ENTER_KEY_CODE,
   SPACE_KEY_CODE,
   ESCAPE_KEY_CODE,
   ARROW_KEY_CODES,
-} from 'ember-sortable/test-support/utils/keyboard';
+} from 'ember-sortable/test-support';
```
NullVoxPopuli, line 361: this change wasn't exactly needed, but it makes controlling the scope of "Public API" easier in the package.json exports.
additionally, having a single export for test-support is helpful for consumers discovering what even is available to use for testing (thanks to LSPs)

addon/package.json
```diff
@@ -35,7 +37,8 @@
   "peerDependencies": {
     "ember-modifier": "^3.2.0 || ^4.0.0",
     "@ember/test-helpers": "^2.6.0",
-    "@ember/test-waiters": "^3.0.1"
+    "@ember/test-waiters": "^3.0.1",
+    "ember-source": "^3.28.0 || ^4.0.0"
```
NullVoxPopuli, line 41: ember-source needs to be a peer as we import things from ember-source

addon/package.json
```diff
@@ -55,11 +59,28 @@
...
     "prettier": "^2.5.1",
-    "rollup": "^3.10.0"
+    "release-it": "^15.6.0",
+    "rollup": "^3.10.0",
+    "rollup-plugin-copy": "^3.4.0"
   },
   "engines": {
     "node": "14.* || >= 16"
   },
+  "release-it": {
```
NullVoxPopuli, line 69: moved from the monorepo-root.

release-it doesn't need to support monorepos, because this one workspace is the only thing published to npm

addon/rollup.config.mjs
```diff
@@ -1,4 +1,5 @@
 import babel from '@rollup/plugin-babel';
+import copy from 'rollup-plugin-copy';
```
NullVoxPopuli, line 2: sync's with the v2 addon-blueprint

addon/rollup.config.mjs
```diff
@@ -14,7 +15,7 @@ export default {
   plugins: [
     // These are the modules that users should be able to import from your
     // addon. Anything not listed here may get optimized away.
-    addon.publicEntrypoints(['modifiers/**/*.js', 'services/**/*.js', 'test-support/**/*.js']),
+    addon.publicEntrypoints(['**/*.js']),
```
NullVoxPopuli, line 18: generally better sourcemap output occurs when publicEntrypoints is all files.

I have a task here: https://github.com/embroider-build/embroider/issues/1329
that would be a breaking change for `@embroider/addon-dev`, but it will make managing package.json#exports much easier, and optimize sourcemap generation as well as not pre-emptively optimize assets.

Rollup's default behaviors are meant for more production-ish builds, and since addons don't emit production-ready assets (we want the consuming app to make all those decisions), we want to be a bit lazier here for the dist-output. 
This strategy also makes patching and editing `node_modules` quite a bit nicer, if someone needs to do that.

docs/package.json
```diff
@@ -1,17 +1,11 @@
 {
   "name": "docs",
-  "version": "5.0.0",
```
NullVoxPopuli, line 3: private workspaces don't need some fields that were specified in here

docs/package.json
```diff
@@ -23,7 +17,11 @@
     "test:ember": "ember test",
     "test:all": "ember try:each"
   },
+  "dependencies": {
+    "ember-sortable": "workspace:../addon"
```
NullVoxPopuli, line 21: this explicitly declares which ember-sortable we want to use (always local workspace)

test-app/package.json
```diff
@@ -22,12 +16,22 @@
...
   },
-  "devDependencies": {
-    "@ember/optional-features": "^2.0.0",
+  "dependencies": {
     "@ember/test-helpers": "^2.6.0",
     "@ember/test-waiters": "^3.0.1",
+    "ember-modifier": "^3.2.0",
+    "ember-sortable": "workspace:../addon"
+  },
+  "dependenciesMeta": {
```
NullVoxPopuli, line 25: because we need ember-sortable to behave _like a real dependency_ we need this injected setting. This allows the local workspace of ember-sortable to be treated as if it were actually from npm -- there is an open issue on the pnpm repo about making this _default_, because the behavior is most correct

NullVoxPopuli:
I've confirmed that `release-it --dry-run` works, so I'm reasonably confident I can run a release after merge.

(I'll also update the changelog to include everything in this PR)

knownasilya (approved):
Nice. Learned some things along the way too

rwwagner90 (approved):
LGTM! 👍

Issue #520, opened 2023-11-14
Drag is translating `margin-top` upward on other sortable-items


KeithKosh:
**Describe the bug**
I have a draggable list that for the most part is uniform, but some of the sortable-items have different content and have margin-top in them to match the screen designs. When I begin to drag any item without margin-top, any sortable-item that does have margin-top is translated upward by the margin-top value.

**To Reproduce**
Create any sortable-list with an item that has margin-top in it (from my testing it doesn't matter if the style is right on the sortable-item or even on a child DIV inside the sortable-item). Begin dragging an item that does not have margin top. As soon as the drag starts, the rows with margin-top are translated upwards. You can see this if you look in the element inspector, as those rows have a `transform: translateY()` style applied to them.

Also can be reproduced by going to the [demo page](https://adopted-ember-addons.github.io/ember-sortable/demo/) and adding margin-top to any row in the first draggable, then dragging any of the other rows (so for example, adding margin to "Dos" and starting to drag "Uno").

**Expected behavior**
Starting a drag should not cause other items in the list to jump position.

PR #560, opened 2024-06-11
Add direction grid

mkszepp:
This feature allows users to create a sortable list that works with a grid. Items can be moved up, down, right, or left within the same list.

### Motivation?
We are currently using [ember-drag-sort](https://github.com/lolmaus/ember-drag-sort), which supports this feature. Unfortunately, the addon uses classic components, and updating it to the latest edition would require significant effort. Additionally, it is not as flexible as this addon. The drag-and-drop animation in ember-drag-sort is a simple arrow with no options for customization.

Implementing this feature in `ember-sortable` will help us transition away from ember-drag-sort. This update might also encourage others still using the previous addon to switch to this one.

grafik

### DIFF
--- a/addon/src/modifiers/sortable-item.js
+++ b/addon/src/modifiers/sortable-item.js
@@ -865,9 +865,16 @@ export default class SortableItemModifier extends Modifier {
   get height() {
     let el = this.element;
     let height = el.offsetHeight;
+    let elStyles = getComputedStyle(el);
 
-    let marginBottom = parseFloat(getComputedStyle(el).marginBottom);
-    height += marginBottom;
+    // This is needed atm only for grid, so fix jumping on drag-start.
+    // In test-app it looks like there is a side-effect when we activate also for direction vertical.
+    // If any user will anytime report a jumping in vertical direction, just remove activate for every direction
+    if (this.direction === 'grid') {
+      height += parseFloat(elStyles.marginTop);
+    }
+
+    height += parseFloat(elStyles.marginBottom);
 
     height += getBorderSpacing(el).vertical;
 
--- a/addon/src/modifiers/sortable-item.js
+++ b/addon/src/modifiers/sortable-item.js
@@ -181,6 +181,46 @@ export default class SortableItemModifier extends Modifier {
   }
 
   /**
+   * Gives info in which direction the element will be moved
+   *
+   * @property moveDirection
+   * @type Object
+   */
+  get moveDirection() {
+    const moveDirection = {
+      left: false,
+      right: false,
+      top: false,
+      bottom: false,
+    };
+
+    if (!this.isDragging) {
+      return moveDirection;
+    }
+
+    const dragOriginX = this._dragOriginX;
+    const dragOriginY = this._dragOriginY;
+
+    if (dragOriginX > this._pageX) {
+      moveDirection.left = true;
+    }
+
+    if (dragOriginX < this._pageX) {
+      moveDirection.right = true;
+    }
+
+    if (dragOriginY > this._pageY) {
+      moveDirection.top = true;
+    }
+
+    if (dragOriginY < this._pageY) {
+      moveDirection.bottom = true;
+    }
+
+    return moveDirection;
+  }
+
+  /**
    Action that fires when the item starts being dragged.
    @property onDragStart
    @type Function
@@ -440,7 +480,7 @@ export default class SortableItemModifier extends Modifier {
     };
 
     let leadingEdgeKey, trailingEdgeKey, scrollKey, pageKey;
-    if (groupDirection === 'x') {
+    if (groupDirection === 'grid' || groupDirection === 'x') {
       leadingEdgeKey = 'left';
       trailingEdgeKey = 'right';
       scrollKey = 'scrollLeft';
@@ -516,6 +556,33 @@ export default class SortableItemModifier extends Modifier {
     let scrollOrigin;
     let parentElement = this.element.parentNode;
 
+    if (groupDirection === 'grid') {
+      this.startEvent = startEvent;
+      const dragOriginX = getX(startEvent);
+      this._dragOriginX = getX(startEvent);
+      const elementOriginX = this.x;
+      const scrollOriginX = parentElement.getBoundingClientRect().left;
+
+      const dragOriginY = getY(startEvent);
+      this._dragOriginY = dragOriginY;
+      const elementOriginY = this.y;
+      const scrollOriginY = parentElement.getBoundingClientRect().top;
+
+      return (event) => {
+        this._pageX = getX(event);
+        let dx = this._pageX - dragOriginX;
+        let scrollX = parentElement.getBoundingClientRect().left;
+        let x = elementOriginX + dx + (scrollOriginX - scrollX);
+
+        this._pageY = getY(event);
+        let dy = this._pageY - dragOriginY;
+        let scrollY = parentElement.getBoundingClientRect().top;
+        let y = elementOriginY + dy + (scrollOriginY - scrollY);
+
+        this._drag(x, y);
+      };
+    }
+
     if (groupDirection === 'x') {
       dragOrigin = getX(startEvent);
       elementOrigin = this.x;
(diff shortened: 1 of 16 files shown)


mkszepp:
@NullVoxPopuli can you make a rerun of test? Local it is passing without any issue and also same changes on my repo are getting a green cli: https://github.com/mkszepp/ember-sortable/pull/1

mkszepp:
@NullVoxPopuli do you have any plan when we can get this feature?

NullVoxPopuli:
Sorry, my email situation is a bit wonky atm <3 feature looks good -- gonna review the code

addon/src/modifiers/sortable-item.js
```diff
@@ -614,10 +685,15 @@ export default class SortableItemModifier extends Modifier {
 
     this._preventClick();
 
+    // Get sortableItems before making drop end and pass it to update function
```
NullVoxPopuli, line 688: thanks for the comment!

docs/ember-cli-build.js
```diff
@@ -5,7 +5,7 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 module.exports = function (defaults) {
   let app = new EmberApp(defaults, {
     autoImport: {
-      watchDependencies: ['ember-keyboard'],
+      watchDependencies: ['ember-sortable'],
```
NullVoxPopuli, line 8: thank you!!

NullVoxPopuli (changes requested):
looks good, but can we get some automated tests added? thanks!!

mkszepp:
@NullVoxPopuli test were now added... while adding tests i have discovered two bugs, but now they are fixed

mkszepp:
It looks like the test fails, cause a timing... locally i have got also sometimes a fail... maybe we need to add a short delay in `reorder` test helper... on my fork PR tests are looking good (but i have also needed some test retries)

NullVoxPopuli:
seems like things are failing pretty regularly -- def can't merge without resolving 🤔 

Do you think a test waiter needs to be added so that the waiter system "just works", and consumers of this library then also don't need to worry about adding random waiting in their tests?

mkszepp:
agree, atm its not mergable, beacause tests do fail... i will look to find a fix... maybe `waitUntil` helps me... i will try

addon/src/test-support/helpers/drag.js
```diff
@@ -108,4 +108,11 @@ export async function drag(mode, itemSelector, offsetFn, callbacks = {}) {
     await callbacks.dragend();
     await settled();
   }
+
```
NullVoxPopuli, line 111: you could also try 

```
await new Promise(requestAnimationFrame)
await settled();
```
if there is an un-instrumented request animation frame in this library.


but it seems like we want to start a waiter when we start dragging and then "endAsync" when we stop dragging.

like:

```js
import { buildWaiter } from '@ember/test-waiters';


const waiter = buildWaiter('ember-sortable:dragging');

// ...

// dragstart:
this.waiter = waiter.beginAsync();

// dragend:
waiter.endAsync(this.waiter);

```

or something like that -- then arbitrary waiting would never need to happen _and_ if something goes wrong with the dragging, we have tools to find out exactly where:
- getPendingWaiterState()
- getSettledState()

(I usually assign both of those to `window` in my test-helper.js, so I can invoke them from the console if I run in to a hanging test)

addon/src/modifiers/sortable-item.js
```diff
@@ -706,7 +824,8 @@ export default class SortableItemModifier extends Modifier {
    @type Number
    */
   get transitionDuration() {
-    let el = this.element;
+    const items = this.sortableGroup.sortedItems.filter((x) => !x.isDragging && !x.isDropping);
+    let el = items[0].element ?? this.element; // Fallback when only one element is present in list
```
mkszepp, line 828: This was since now not correct... when someone has set transition to 1 sec or more, vertical and also horiziontal elements were jumping, because the duration of moved element is always "0"

addon/src/test-support/helpers/drag.js
```diff
@@ -108,4 +108,11 @@ export async function drag(mode, itemSelector, offsetFn, callbacks = {}) {
...
+  await waitUntil(
+    () => {
+      return !find('.is-dropping');
+    },
+    { timeout: 2000 }
+  );
```
mkszepp, line 117: This helps us to fix tests, because in tests there is no point in which we do wait for `transitionEnd`. This bug is possible to reproduce, when transition time is high or when someone has a lot of items

addon/src/modifiers/sortable-item.js
```diff
@@ -366,6 +406,9 @@ export default class SortableItemModifier extends Modifier {
    * @private
    */
   _prepareDrag(startEvent, event) {
+    if (this.sortableGroup.sortedItems.some((x) => x.isBusy)) {
+      return;
+    }
```
mkszepp, line 411: always check if any item is busy, otherwise lock dragging. There is a isBusy check in an other place, but it checks always one the current item which you want drag.
Without this check can already drag items, without finishing the transition from old element. Simple repro is, set transition time to 1 sec

addon/src/modifiers/sortable-item.js
```diff
@@ -614,11 +688,19 @@ export default class SortableItemModifier extends Modifier {
...
     set(this, 'isDragging', false);
     set(this, 'isDropping', true);
 
-    this.sortableGroup.update();
-    transitionPromise.then(() => this._complete());
+    this.sortableGroup.update(sortedItems);
+
+    let allTransitionPromise = this._waitForAllTransitions();
+
+    Promise.all([transitionPromise, allTransitionPromise]).then(() => this._complete());
```
mkszepp, line 703: `_waitForAllTransitions()` is internally checking if any element is still in transition. Thats important, because setting complete must be success after all items did complete transition (not only like now, dragged element).
Without this check, the complete will fired early and this is causing randomly errors (like we have now got with direction grid)

mkszepp:
@NullVoxPopuli i have left some comments... all test should now pass (without retries 😅)

While adding this feature i was running in errors, which are still present in current version, but tests were never failed beacuse the lists were always small.
`drag` has never waited for `transitionEnd`, it works with 5 elements without any issue, but when we use more (can' t tell you the exact number) this tests do always randomly fail like we have now got in grid.
For consumer apps this change should not be braking

There was also never waited if all transition from items were completed, by using transition time with 1 secound + you will get crazy animations 😂. With my fix now we do wait since everything is completed.

addon/src/modifiers/sortable-item.js
mkszepp: this changes were moved to PR #562

addon/src/modifiers/sortable-item.js
mkszepp: this change was moved to PR #562

addon/src/test-support/helpers/drag.js
mkszepp: this change was moved to PR #562

addon/src/modifiers/sortable-item.js
mkszepp: this change was moved to PR #562

addon/src/modifiers/sortable-item.js
```diff
@@ -833,9 +909,16 @@ export default class SortableItemModifier extends Modifier {
   get height() {
     let el = this.element;
     let height = el.offsetHeight;
+    let elStyles = getComputedStyle(el);
+
+    // This is needed atm only for grid, to fix jumping on drag-start.
+    // In test-app it looks like there is a side-effect when we activate also for direction vertical.
+    // If any user will anytime report a jumping in vertical direction, we should activate for every direction and fix our test-app
+    if (this.direction === 'grid') {
+      height += parseFloat(elStyles.marginTop);
```
mkszepp, line 918: I have looked shortly in issues.. i think generally adding this could be fix this issue #520, but this should be done in next PR

NullVoxPopuli: fractional pixels can be a doozy!

NullVoxPopuli (approved):

PR #565, opened 2024-07-03
Add TypeScript type definitions

### LINKED ISSUE #564: Addon doesn't supply TypeScript type definitions

charlesfries:
Should fix #564

### DIFF
--- a/addon/declarations/modifiers/sortable-group.d.ts
+++ b/addon/declarations/modifiers/sortable-group.d.ts
@@ -1,0 +1,13 @@
+declare module 'ember-sortable/modifiers/sortable-group' {
+  import Modifier from 'ember-modifier';
+
+  export default class SortableGroup<T> extends Modifier<{
+    Args: {
+      Named: {
+        groupName?: string;
+        onChange: (value: T[]) => void;
+      };
+    };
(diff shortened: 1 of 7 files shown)


NullVoxPopuli:
any chance you can add type-checking to CI within the test-app?

the test-app could just have one TS file that is type-checked -- no need to convert the whole test-app to TS

charlesfries:
@NullVoxPopuli One point of clarification--do you want me to import these definitions I've added in this PR into that TS file, or just convert some existing file to TS?

NullVoxPopuli:
it can be a standalone file -- no need to convert existing stuff.

like, you could have `app/type-tests` in the test-app, and have scenarios specifically for ensuring the APIs are correct -- no need to have runnable code as `glint` would verify that the types are correct.

> , which necessitates upgrading ember-source to a later version that includes the built in TS types. 

we could also do a separate project, maybe `test-types` (rather than test-app), which is the most minimal set of files we need to type check -- could even use built in types, and you don't need most of the files that come with an app

mkszepp:
converted the complete addon to TS #581 so this PR is not anymore needed... btw... the signatures here are not complety

PR #566, opened 2024-07-12
Fix grid direction, when group element has negative margin

mkszepp:
While using direction grid in our app i have discoverd, that moving to first position of group isn't working.

The bug occures, when the group has a negativ margin on left.
This is caused, because we are using a grid system like bootstrap, which brings on row's negative margins.

To support this, i have updated the grid demo and added a minimal fix.


mkszepp:
Like talked over DM, we are readding old test case

NullVoxPopuli (approved):

PR #569, opened 2024-07-12
Update rootURL for gh-pages

NullVoxPopuli:


mansona:
do we want to deploy this on netlify so that we can have preview urls on PRs?

NullVoxPopuli:
> do we want to deploy this on netlify so that we can have preview urls on PRs?

I'm anti netlify, but if you want to set it up, no objection!

mansona:
> I'm anti netlify, but if you want to set it up, no objection!

So first of all, everyone is entitled to their opinions so I'm not trying to mitigate that. My personal opinion is that Netlify is absolutely fantastic and a really good thing for the web, especially for people getting started and wanting to get their apps in front of people. Netlify have also been very good to the Ember community with their provision of free hosting for all of our docs websites so I always have nothing but positive things to say about Netlify whenever I get a chance.

I understand that you are communicating a personal opinion when you say things like this, but I encourage you to think twice and be careful about what you say. You are a member of the Ember Core Tooling team, and whether you like it or not, your words carry an implicit weight of "I'm a member of the Ember Core team, and I believe ...". 

The Ember Core team is very pro-Netlify and is grateful for everything that they have given Ember over the years 👍

PR #575, opened 2024-07-31
Fix error `TypeError: items[0] is undefined` in `sortable-item` modifier

mkszepp:
In our app we have seen that when we have only one item we are running into error `TypeError: items[0] is undefined` this is caused, because the `filter` returns possible a empty array

The bug was introduced by adding direction `grid` https://github.com/adopted-ember-addons/ember-sortable/pull/560

Workaround in our app (since this bugfix landes) is to add `disabled` on group


NullVoxPopuli (approved):
Is there a way to add a test for this?

mkszepp:
@NullVoxPopuli i think difficulte. The better way should be to move the addon to TS, so the errors are already visibile while developing :)

mkszepp:
@NullVoxPopuli test added

NullVoxPopuli (approved):

PR #576, opened 2024-08-03
Remove `prepublishOnly` in favor of `prepack` script to fix randomly test failing

mkszepp:
More/less always while building addon we are running into the error that tests do fail... `prepublishOnly` runs at same time like `build`.

Removing `prepublishOnly` (which makes the same as build) should fix the randomly fails

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -27,6 +27,7 @@
   ],
   "scripts": {
     "build": "rollup --config",
+    "prepack": "rollup --config",
     "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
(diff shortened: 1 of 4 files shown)


addon/package.json
```diff
@@ -27,7 +27,6 @@
   ],
   "scripts": {
     "build": "rollup --config",
```
NullVoxPopuli, line 29:
```suggestion
    "build": "rollup --config",
    "prepack": "rollup --config",
```

NullVoxPopuli: I think this is what we want instead

NullVoxPopuli (approved):

PR #584, opened 2024-09-25
fix(ts): make onChange argument optional

lukasnys:
In the `addon/src/modifiers/sortable-group.ts` file, the `onChange` argument is marked as required. However, looking at the implementation, it looks like it should be optional instead.


NullVoxPopuli:
What's the usecase for _not_ having `@onChange`?

mkszepp:
i think the only usecase for optional with latest is when there is `disabled=true` or? because without `onChanges` the data will never be changes

Looking into code docs there is declared as `optional`, but maybe its historcally, when there was 2-way binding... is it possible?

https://github.com/adopted-ember-addons/ember-sortable/blob/0e28f8bbfe4311ca25a068d85b5062b485975b08/addon/src/modifiers/sortable-group.ts#L59-L78

PR #587, opened 2024-09-27
fix types for reorder and drag test helpers, fixes #586

### LINKED ISSUE #586: Types or test-support/reorder are not correct
The current type declaration is
```ts
export declare function reorder<T extends keyof HTMLElementTagNameMap>(mode: TMode, itemSelector: string, ...resultSelectors: T[]): Promise<void>;
```

which only allows for HTML Element tag names (such as 'a', 'div', 'li') to be used as resultSelector

The example in the source file:
image

also throws a type error
image

BoussonKarel:
We can use Parameters<typeof find>[0] because we are passing the selectors to the find function.
```ts
  itemSelector: Parameters<typeof findAll>[0],
  ...resultSelectors: Parameters<typeof find>[0][]
```

or we can just copy their types
```ts
  itemSelector: keyof (HTMLElementTagNameMap | SVGElementTagNameMap) | string,
  ...resultSelectors: (keyof (HTMLElementTagNameMap | SVGElementTagNameMap) | string)[]
```

What's the best / cleanest / preferred method of doing this?

### DIFF
--- a/addon/src/test-support/helpers/drag.ts
+++ b/addon/src/test-support/helpers/drag.ts
@@ -37,7 +37,7 @@ interface Callbacks {
 
 export async function drag(
   mode: TMode,
-  itemSelector: string,
+  itemSelector: keyof (HTMLElementTagNameMap | SVGElementTagNameMap) | string, // or Parameters<typeof find>[0][]
   offsetFn: () => { dx: number; dy: number },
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):
This seems fine. If we want to change something, we can do another release!

PR #600, opened 2024-11-29
`firstItemPosition` computed property is not recomputed when `sortedItems` changes

gwak:
This PR proposes a fox for the test case highlighted in the PR https://github.com/adopted-ember-addons/ember-sortable/pull/599.


gwak:
Hi @cyril-sf, can any maintainer look into this [issue](https://github.com/adopted-ember-addons/ember-sortable/pull/599) and related PR ? We're  currently using our fork in our main production application, but would prefer not to...

NullVoxPopuli:
The changes to the test app feel a smidge excessive? Why'd every example have to change? Can something more targeted happen instead?

NullVoxPopuli:
Also, biggest apologies for not commenting sooner. I don't know how i missed this PR <3

PR #606, opened 2025-01-10
bug: Support when border-spacing returns a single value

cyril-sf:
This is mostly a workaround for a potential bug introduced with Firefox 134

Based on the spec, `getComputedStyle(el).borderSpacing` should always return 2 values, even when they are identical.

On Firefox 134, it returns a single value when they are the same.


NullVoxPopuli (approved):

cyril-sf:
@NullVoxPopuli lint fixed!

NullVoxPopuli (approved):

PR #677, opened 2026-08-05
fix: don't abort drop when a drop animation is cancelled (#641)

### LINKED ISSUE #641: AbortError thrown on fast drag/drop
**Describe the bug**
When quickly dragging and dropping an item an AbortError is thrown which breaks further drag/drop functionality.

**To Reproduce**

1. Using <ul {{sortable-group}}> with {{#each items as |item|}} <li {{sortable-item model=item}}>
2. Drag an item quickly and release before animation completes
3. Observe the console error:
```
AbortError: the user aborted a request.
```

**Expected behavior**
Transitions should gracefully cancel or resolve. No uncaught errors should be thrown.

**Screenshots**

Image

**Additional context**

- The DOM node is not removed
- The .is-dropping class is applied
- Transition seems to be interrupted mid-flight
- Seeing in Chrome and Safari
- Happens with both ember-sortable implementations we have in the codebase
- Patching [sortable-item.js](https://github.com/adopted-ember-addons/ember-sortable/blob/3c9c50bb742e784ad2c045451e3cf7fe99102217/addon/src/modifiers/sortable-item.ts#L884) resolves the issue for me locally

```
      return transitionPromise;
    } catch (e) {
      if (e.name === 'AbortError') {
        return Promise.resolve();
      }
      throw e;
    }
```

**Versions**
ember-sortable: 5.3.3
ember: 4.4

samridhivig:
## What / Why

Fixes #641.

On drop, `_waitForAllTransitions` awaits each item's Web Animations `finished` promise:

```js
const animationPromises = animations.flatMap((animationList) => {
  return animationList.map((animation) => animation.finished);
});
transitionPromise = Promise.all(animationPromises);
```

When a transition is interrupted mid-drop — e.g. by the re-render triggered from `sortableGroup.update`, or a fast drag/drop before the animation finishes — the browser **cancels** the animation, and per the Web Animations spec `animation.finished` rejects with a `DOMException` named `AbortError` ("The user aborted a request").

`_drop` chains the result with `.then(() => this._complete())` and no rejection handler:

```js
Promise.all([transitionPromise, allTransitionPromise]).then(() => this._complete());
```

So the AbortError becomes an **uncaught promise rejection**, `_complete()` never runs, and `onDragStop` / the group's `onChange` never fires. The item keeps its `.is-dropping` class and **all subsequent drag/drop is broken**. Reproducible in Chrome, Firefox and Safari (matches the reports in #641).

## Fix

Treat a cancelled animation as finished (a cancelled animation rejecting `finished` is expected per spec), so the drop still commits. Any other rejection is re-thrown so genuine failures aren't masked:

```js
animation.finished.catch((error: unknown) => {
  if (error instanceof DOMException && error.name === 'AbortError') {
    return;
  }
  throw error;
})
```

## Testing

Manually verified against the reproduction in #641: with the fix, interrupting a drop no longer throws, the reorder commits, and subsequent drags keep working. Happy to add an automated test if maintainers can point me at the best way to simulate a cancelled `getAnimations()` transition in the test app.

### DIFF
--- a/addon/src/modifiers/sortable-item.ts
+++ b/addon/src/modifiers/sortable-item.ts
@@ -868,7 +868,19 @@ export default class SortableItemModifier<T> extends Modifier<SortableItemModifi
       const animations = this.sortableGroup.sortedItems.map((x) => x.element.getAnimations());
 
       const animationPromises = animations.flatMap((animationList) => {
-        return animationList.map((animation) => animation.finished);
+        return animationList.map((animation) =>
+          // A cancelled animation (e.g. one interrupted by the re-render that
+          // happens mid-drop) rejects `finished` with an AbortError. Treat it
+          // as finished so the drop still completes, instead of leaving this
+          // promise chain unhandled and blocking all subsequent drag/drop.
+          // Re-throw anything else. See #641.
+          animation.finished.catch((error: unknown) => {
+            if (error instanceof DOMException && error.name === 'AbortError') {
+              return;
+            }
+            throw error;
+          }),
+        );
       });
 
       transitionPromise = Promise.all(animationPromises);
--- a/test-app/tests/integration/modifiers/sortable-item-test.js
+++ b/test-app/tests/integration/modifiers/sortable-item-test.js
@@ -8,6 +8,61 @@ import { hbs } from 'ember-cli-htmlbars';
 module('Integration | Modifier | sortable-item', function (hooks) {
   setupRenderingTest(hooks);
 
+  // #641: on drop, `_waitForAllTransitions` awaits each item's Web Animations
+  // `finished` promise. If an animation gets cancelled mid-drop, say by the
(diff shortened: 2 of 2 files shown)


knownasilya:
@samridhivig can you write a test for this?

samridhivig:
> @samridhivig can you write a test for this?

added a test case

BoussonKarel:
Confirming this fixes the flaky "AbortError: The user aborted a request" failures we're seeing on master's CI (e.g. https://github.com/adopted-ember-addons/ember-sortable/actions/runs/32402402070/job/96534062089?pr=680, on `smoke modifier: reordering with mouse/touch events grid` and `grid demo 2` — those tests do many rapid successive drags, which is exactly what cancels the drop animation and triggers the unhandled `AbortError` here.

Would be great to get this merged to stabilize CI.

knownasilya:
@samridhivig could you fix the linting errors?

